PR #24, opened 2025-02-17
Sync Tempo Config with Compose File

shonarik:
While there are definitely other solutions, the selected solution copies the `tempo` source:
https://github.com/grafana/tempo/tree/main/example/docker-compose/local

The paths in the compose file and the tempo config have been aligned.


gearnode:
@shonarik Thanks a lot 🎉

PR #149, opened 2025-06-12
Add pages to the new front end

SachaProbo:
Add new pages from the old front end to the new.


apps/console2/src/hooks/graph/AssetGraph.ts
```diff
@@ -0,0 +1,276 @@
...
+                    name
+                    websiteUrl
+                  }
+                }
+              }
+              createdAt
+            }
+          }
+        }
+        peoples(first: 100) {
```
gearnode: use people selector instead of doing this query

PR #192, opened 2025-07-16
Reset controls framework foreign key

SachaProbo:
## Summary by cubic
Reset the foreign key on the controls table to ensure all controls reference existing frameworks and to enable cascading deletes.

- **Migration**
  - Removes invalid controls with missing frameworks.
  - Re-adds the foreign key with ON DELETE CASCADE.


gearnode (approved):
LGTM

PR #200, opened 2025-07-18
Exclude controls

SachaProbo:
## UI:

Screenshot 2025-07-18 at 17 36 28
Screenshot 2025-07-18 at 17 36 39
Screenshot 2025-07-18 at 17 37 55

    

---

## Summary by cubic
Added support for excluding controls from frameworks, including a status field and exclusion justification for each control.

- **New Features**
  - Controls can now be marked as "excluded" with a required justification.
  - Excluded controls are visually indicated in the UI and included in exports.
  - API, database, and forms updated to support control status and justification.

### DIFF
--- a/packages/ui/src/Atoms/ControlItem/ControlItem.tsx
+++ b/packages/ui/src/Atoms/ControlItem/ControlItem.tsx
@@ -6,6 +6,7 @@ type Props = {
     active?: boolean;
     id: string;
     description?: string;
+    excluded?: boolean;
     to: string;
 } & HTMLAttributes<HTMLAnchorElement>;
 
@@ -26,19 +27,39 @@ const classNames = tv({
                 id: "bg-highlight",
             },
         },
+        excluded: {
+            true: {
+                wrapper: "opacity-60",
+            },
+            false: {
+                wrapper: "",
+            },
+        },
     },
+    compoundVariants: [
+        {
+            active: true,
+            excluded: true,
+            class: {
+                wrapper: "bg-tertiary-pressed opacity-60",
+                id: "bg-active",
+            },
+        },
+    ],
     defaultVariants: {
         active: false,
+        excluded: false,
     },
 });
 
-export function ControlItem({ active, id, description, to, ...props }: Props) {
+export function ControlItem({ active, id, description, excluded, to, ...props }: Props) {
     const {
         wrapper,
         id: idCls,
         description: descriptionCls,
     } = classNames({
         active,
+        excluded,
     });
 
     const ref = useRef<HTMLAnchorElement>(null);
--- a/pkg/coredata/control.go
+++ b/pkg/coredata/control.go
@@ -28,22 +28,26 @@ import (
 
 type (
 	Control struct {
-		ID           gid.GID      `db:"id"`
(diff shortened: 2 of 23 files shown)


pkg/coredata/control.go
```diff
@@ -747,15 +781,19 @@ RETURNING
...
 `
 	q = fmt.Sprintf(q, scope.SQLFragment())
 
 	args := pgx.StrictNamedArgs{
-		"control_id":    c.ID,
-		"section_title": params.SectionTitle,
-		"updated_at":    time.Now(),
+		"control_id":              c.ID,
+		"section_title":           params.SectionTitle,
+		"status":                  params.Status,
```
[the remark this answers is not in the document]
SachaProbo: There is a check after.

pkg/probo/framework_service.go
```diff
@@ -306,37 +307,54 @@ func (s FrameworkService) StateOfApplicability(ctx context.Context, frameworkID
 			}
 
 			for _, control := range controls {
+				exclusionJustification := ""
+				if control.ExclusionJustification != nil && control.Status == coredata.ControlStatusExcluded {
```
gearnode: I would only use `control.Status == coredata.ControlStatusExcluded` this as a condition to avoid not expected behavior in the future.

pkg/probo/framework_service.go
```diff
@@ -306,37 +307,54 @@ func (s FrameworkService) StateOfApplicability(ctx context.Context, frameworkID
...
-						page.OrderBy[coredata.MeasureOrderField]{
-							Field:     coredata.MeasureOrderFieldCreatedAt,
-							Direction: page.OrderDirectionAsc,
-						},
-					),
-					coredata.NewMeasureFilter(nil),
-				)
-				if err != nil {
-					return fmt.Errorf("cannot load measures: %w", err)
+				if control.Status == coredata.ControlStatusIncluded {
```
gearnode: It will be simpler like that because we need skip all other processing in case of exclusion for this control

f control.Status == coredata.ControlStatusExcluded { 
      ... 
      continue
 }```

packages/ui/src/Atoms/ControlItem/ControlItem.tsx
```diff
@@ -26,19 +27,39 @@ const classNames = tv({
                 id: "bg-highlight",
             },
         },
+        excluded: {
```
gearnode, line 30: the active false style does not work?

pkg/probo/framework_service.go
SachaProbo: The append row is in the loop at the end so  I can't just exit the loop. I could refacto the loop but I am not sure it will be more clear this way.

packages/ui/src/Atoms/ControlItem/ControlItem.tsx
SachaProbo: What do you mean?

pkg/probo/framework_service.go
SachaProbo: I did the change. Still keep the check but it raise an error in this case.

Issue #222, opened 2025-08-04
Replace `fmt.Sprintf` email templates with Go `text/template` embedded templates


gearnode:
Currently, email bodies are constructed using `fmt.Sprintf` with plain string variables. We plan to migrate to Go’s `text/template` embedded templating system to improve maintainability, readability, and safety of email content formatting.

rhythmp17:
Hi @gearnode, please review my pull request.

PR #276, opened 2025-08-17
Use text/template instead of fmt.Spritf for email template. Signed-off-by: Author Name <rhythmpatni17@gmail.com>

rhythmp17:
This PR is a fix for issue: #222 
    

---

## Summary by cubic
Switched signing request emails to use Go’s text/template instead of fmt.Sprintf. This centralizes the email copy and makes it safer and easier to change.

- **Refactors**
  - Added signEmailTemplate and SignEmailData for email rendering.
  - Replaced inline fmt.Sprintf with template execution to build TextBody.
  - Updated email copy to include a closing line (“Regards, Probo Team”).


pkg/probo/document_service.go
```diff
@@ -435,12 +447,26 @@ func (s *DocumentService) SendSigningNotifications(
 					}.Encode(),
 				}
 
+				tmpl, err := template.New("signRequest").Parse(signEmailTemplate)
```
gearnode, line 450: We want use the following pattern in a variable on top of the file: `var t = template.Must(template.New("name").Parse("html"))`

gearnode: We want use the following pattern in a variable on top of the file: `var t = template.Must(template.New("name").Parse("html"))`

This ensure we are not able to build the application if the template has syntax error.

pkg/probo/document_service.go
```diff
@@ -435,12 +447,26 @@ func (s *DocumentService) SendSigningNotifications(
...
+						panic(err)
+				}
+
+				data := SignEmailData{
+						SignURL: signRequestURL.String(),
+				}
+
+				var body bytes.Buffer
+				if err := tmpl.Execute(&body, data); err != nil {
+						panic(err)
```
gearnode: We should not use panic here, instead we have to return a wrapped error. In that layer we never use any panic in runtime code.

gearnode (changes requested):
Thanks for your contribution! Small changes before be able to merge it!

pkg/probo/document_service.go
rhythmp17: Made this change.

pkg/probo/document_service.go
rhythmp17: Resolved.

rhythmp17:
Hi @gearnode , I have made the changes according to the feedback received.

PR #299, opened 2025-08-21
Redirect user to login page after password reset

BrenoNunes96:
This PR implements a fix for the password reset flow. Now, after a user successfully resets their password, they are automatically redirected to the login page. 

Previously, the user remained on the password reset page, which could cause confusion. This change improves the user experience by guiding them directly to the login page to access their account with the new password.

    

---

## Summary by cubic
Automatically redirect users to the login page after a successful password reset so they can sign in with the new password. This fixes the previous behavior where users stayed on the reset page.

### DIFF
--- a/apps/console/src/pages/auth/ResetPasswordPage.tsx
+++ b/apps/console/src/pages/auth/ResetPasswordPage.tsx
@@ -74,7 +74,7 @@ export default function ResetPasswordPage() {
       description: __("Password reset successfully"),
       variant: "success",
     });
-    navigate("/auth/login", { replace: true });
+    navigate("/auth/LoginPage.tsx", { replace: true });
   });
 
   usePageTitle(__("Reset password"));


apps/console/src/pages/auth/ResetPasswordPage.tsx
```diff
@@ -74,7 +74,7 @@ export default function ResetPasswordPage() {
       description: __("Password reset successfully"),
       variant: "success",
     });
-    navigate("/auth/login", { replace: true });
+    navigate("/auth/LoginPage.tsx", { replace: true });
```
gearnode, line 77: use `auth/login` instead of the component name

gearnode (changes requested):
Thanks for your contribution, one small change and we'll be able to merge it.

SachaProbo:
Thanks for your contribution. The bug has been fixed in another PR.

PR #364, opened 2025-09-16
Change trufflehog workflow permission

SachaProbo:
## Summary by cubic
Add read-only contents permission to the TruffleHog secrets scan workflow so it runs with a minimal-scope token. This hardens security and avoids unintended write access during scans.


gearnode (approved):
❤️

PR #411, opened 2025-10-02
Add missing s3 metadata

SachaProbo:


pkg/probo/evidence_service.go
```diff
@@ -367,6 +375,10 @@ func (s EvidenceService) UploadMeasureEvidence(
 			Key:         aws.String(objectKey.String()),
 			Body:        req.File.Content,
 			ContentType: aws.String(req.File.ContentType),
+			Metadata: map[string]string{
+				"type":        "evidence",
+				"evidence-id": evidenceID.String(),
```
gearnode: Add organization id, same for all

PR #417, opened 2025-10-03
Handle missing NDA

SachaProbo:
## Summary by cubic
Skip NDA enforcement when a trust center has no NDA configured, and only show the NDA dialog when an NDA file exists. Also persist the NDA file ID on acceptance and make the NDA URL resolver tolerant to missing/unauthenticated cases.

- **Bug Fixes**
  - Frontend: show NDA dialog only when authenticated, not yet accepted, and ndaFileUrl exists.
  - Backend: ExportDocumentPDF/ExportReportPDF enforce NDA only if the trust center has an NDA file; still enforce document/report access.
  - GraphQL: NdaFileURL returns nil on failure or unauthenticated users to avoid breaking views.
  - Data: added nda_file_id to trust_center_accesses and persist it on acceptance; queries and updates updated accordingly.


gearnode (approved):
LGTM

PR #439, opened 2025-10-14
Fix failed ACME block the queue

gearnode:
## Summary by cubic
Stops failed ACME and renewal attempts from blocking the SSL queue by adding retry tracking and automatic stale-attempt recovery. Domains now reset cleanly or fail fast after limited retries.

- **Bug Fixes**
  - Track retries with ssl_retry_count and ssl_last_attempt_at.
  - Provisioning: on failure, increment retry, reset to PENDING; after 3 retries, mark FAILED.
  - Renewal: on failure, increment retry and keep ACTIVE; after 3 retries, mark FAILED.
  - Auto-reset stale attempts: if in PROVISIONING/RENEWING for >4h or last attempt >24h, clear challenge data and set to PENDING.
  - On success, clear challenge data and reset retry counters.
  - Added query to find stale domains and logging around resets and failures.

- **Migration**
  - Adds columns: ssl_retry_count (default 0), ssl_last_attempt_at.
  - Adds partial index on (ssl_last_attempt_at, ssl_status) for PENDING/PROVISIONING/RENEWING.

### DIFF
--- a/pkg/certmanager/provisioner.go
+++ b/pkg/certmanager/provisioner.go
@@ -17,6 +17,7 @@ package certmanager
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/getprobo/probo/pkg/coredata"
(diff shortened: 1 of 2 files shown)


pkg/coredata/migrations/20251014T120000Z.sql
```diff
@@ -0,0 +1,8 @@
+-- Add retry tracking fields for SSL certificate provisioning
+ALTER TABLE custom_domains ADD COLUMN ssl_retry_count INTEGER NOT NULL DEFAULT 0;
```
SachaProbo, line 1: You don't want to drop the default after ?

PR #485, opened 2025-10-23
Fix measure input in new task

SachaProbo:
## Summary by cubic
Fixes the measure selector in the New Task form by replacing Select with Combobox. Selection now persists, shows the correct name, and search works as expected.

- **Bug Fixes**
  - Replaced @probo/ui Select/Option with Combobox/ComboboxItem.
  - onSelect updates the form value and syncs the search text to the chosen measure’s name.
  - Display uses the selected measure’s name to prevent blank input after pick.
  - Added resetValueOnHide to clear stale search when closing.
  - Updated Suspense fallback to a disabled Combobox.


gearnode (approved):
LGTM

PR #491, opened 2025-10-27
feat: add cfg generation on docker run using ENV vars

kermitsxb:
This PR basically add cfg file generation on probo docker container start if it doesn't exist on default or specified path. It can still be mounted as a volume from the host system.

The docker-compose.yml file is a copy of the compose.yml file with the containerized probo service added.

Attention should be paid on the default config path, since we don't run `probod` by default but `exec probod -cfg-file "$CONFIG_FILE" "$@"` where default CONFIG_FILE is `/etc/probod/config.yml`

---
## Summary by cubic
Adds a Docker entrypoint that generates /etc/probod/config.yml from environment variables at container start and runs probod with it. Adds a production Docker Compose stack (compose.prod.yaml) with Postgres, MinIO, and Chrome, plus docs for environment variables and installation.

- **New Features**
  - Entrypoint generates config from ENV if missing; uses mounted config when present. Dockerfile installs entrypoint, creates /etc/probod, and switches ENTRYPOINT.
  - compose.prod.yaml starts Probo with sample ENV and ports; added DOCKER_ENVIRONMENT_VARIABLES.md and updated INSTALLATION.md.

- **Migration**
  - Default config path: /etc/probod/config.yml (override with CONFIG_FILE).
  - Set required secrets: PROBOD_ENCRYPTION_KEY, AUTH_COOKIE_SECRET, AUTH_PASSWORD_PEPPER, TRUST_AUTH_TOKEN_SECRET.
  - Mount a config file to skip generation.
  - probod runs via: exec probod -cfg-file "$CONFIG_FILE" "$@".

<sup>Written for commit 134f94c4c8b67c5d07c8c45c5614fb9f5b9f8977. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/DOCKER_ENVIRONMENT_VARIABLES.md
+++ b/docs/DOCKER_ENVIRONMENT_VARIABLES.md
@@ -1,0 +1,193 @@
+# Environment Variables Reference
+
+This document provides a comprehensive reference for all environment variables used by the Probo entrypoint script to generate the configuration file.
+
+## Configuration File
+
+| Variable | Description | Default Value | Required |
+|----------|-------------|---------------|----------|
+| `CONFIG_FILE` | Path to the configuration file | `/etc/probod/config.yml` | No |
+
+## Observability
+
+### Metrics
+
+| Variable | Description | Default Value | Required |
+|----------|-------------|---------------|----------|
+| `METRICS_ADDR` | Address for Prometheus metrics endpoint | `localhost:8081` | No |
+
+### Tracing
+
+| Variable | Description | Default Value | Required |
+|----------|-------------|---------------|----------|
+| `TRACING_ADDR` | OpenTelemetry collector address for distributed tracing | `localhost:4317` | No |
+| `TRACING_MAX_BATCH_SIZE` | Maximum number of spans to batch before export | `512` | No |
+| `TRACING_BATCH_TIMEOUT` | Timeout in seconds for batching spans | `5` | No |
+| `TRACING_EXPORT_TIMEOUT` | Timeout in seconds for exporting traces | `30` | No |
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,185 @@
+services:
```
gearnode, line 1: Is there a reason to have another docker compose file?

docker-compose.yaml
```diff
@@ -0,0 +1,185 @@
...
+  probo:
+    image: "ghcr.io/getprobo/probo:latest"
+    environment:
+      # Required secrets (use secure values in production)
+      PROBOD_ENCRYPTION_KEY: "thisisnotasecretAAAAAAAAAAAAAAAAAAAAAAAAAAA="
```
gearnode, line 6: @SachaProbo we must make the application crash instead of default value in probod, it will be more secure and avoid people running with default value.

@kermitsxb could you in case the key does not exist generate one? Or it's better to keep just default one and ask people to set it properly?

kermitsxb: @gearnode I guess it's better to ask people to set it properly, they'll maybe struggle to find it in the config file afterwards.

docker-compose.yaml
kermitsxb: Good question, currently you're using compose.yaml in dev purpose I reckon ? With probod not running in docker.

That one file is better for self-host purpose. Maybe we should name them differently ? like docker-compose.dev.yml and docker-compose.yml ?

gearnode: Would we move it in contrib folder where we can put all deployment related thing in the future?

kermitsxb: Sure, good idea !

PR #499, opened 2025-10-31
Fix organization creation

SachaProbo:
## Summary by cubic
Fixes organization creation by properly handling the add-member step and safely updating tenant access in the request context. This prevents nil panics and ensures the creator immediately has access to the new org.

- **Bug Fixes**
  - Capture the error from authz.AddUserToOrganization so failures surface.
  - Read userTenantAccess from context and append the new tenant ID only when present, avoiding nil dereferences and keeping access in sync.

<sup>Written for commit a22c06002780a6e9d6eabdb8512b819cc6fd3c2b. Summary will update automatically on new commits.</sup>


gearnode (approved):
LGTM

PR #515, opened 2025-11-07
Fix deletion freeze

SachaProbo:
## Summary by cubic
Fixes the deletion dialog freeze by awaiting the async confirm action and resetting the loading state. The dialog now closes after the confirm finishes, even on errors, to prevent hangs.

<sup>Written for commit b25d30d3c700ed1436aa03b6b91dca2a5efca9d4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/ui/src/Molecules/Dialog/ConfirmDialog.tsx
+++ b/packages/ui/src/Molecules/Dialog/ConfirmDialog.tsx
@@ -79,13 +79,14 @@ export function ConfirmDialog() {
         footer,
     } = dialog();
     const [loading, setLoading] = useState(false);
-    const handleConfirm = () => {
+    const handleConfirm = async () => {
         setLoading(true);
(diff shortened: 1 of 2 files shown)


packages/ui/src/Molecules/Dialog/ConfirmDialog.tsx
```diff
@@ -79,13 +79,16 @@ export function ConfirmDialog() {
...
         setLoading(true);
-        onConfirm()
-            .finally(() => {
-                close();
-                setLoading(false);
-            });
+        try {
+            await onConfirm();
+            close();
+        } catch {
```
gearnode: should we throw something here?

SachaProbo: I removed the empty catch

PR #516, opened 2025-11-09
fix(certmanager#514): check DNS configuration before starting ACME

MustafaAamir:
`checkDNSConfiguration` verifies that the provided domain resolves to cname-target provided in the configuration. Skips if no domain is provided. Returns an error and logs a message if it doesn't resolve.

---
## Summary by cubic
Verify the domain’s CNAME resolves to the configured target before starting ACME HTTP challenges to prevent failed certificate provisioning. Skips ACME and logs a warning when DNS is not ready (addresses certmanager#514).

- **Bug Fixes**
  - Added DNS check in Provisioner to compare the domain’s CNAME to cfg.CustomDomains.CnameTarget.
  - Provisioner now accepts cnameTarget; probod passes it from config.
  - ACME challenge starts only after DNS is verified; otherwise returns a clear error.

<sup>Written for commit 24abdb9d1420dc841e1186a78ec174b44921ed12. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pkg/certmanager/provisioner.go
+++ b/pkg/certmanager/provisioner.go
@@ -17,6 +17,7 @@ package certmanager
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 	"time"
 
@@ -79,9 +80,6 @@ func (p *Provisioner) Run(ctx context.Context) error {
 }
 
 func (p *Provisioner) checkDNSConfiguration(domain string) error {
-	if p.cnameTarget == "" {
-		return nil
-	}
 	cnameRecords, err := net.LookupCNAME(domain)
 	if err != nil {
 		return fmt.Errorf("DNS lookup Failed: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/certmanager/provisioner.go
```diff
@@ -75,6 +78,28 @@ func (p *Provisioner) Run(ctx context.Context) error {
 	}
 }
 
+func (p *Provisioner) checkDNSConfiguration(domain string) error {
+	if p.cnameTarget == "" {
```
gearnode: This error should never occur. If we need to handle an empty case for the CNAME target, it should be done within the `probod` package at the configuration level. But for now we didn't validate the config, we can safely remove this condition.

gearnode:
@MustafaAamir Thanks for your contribution that exactly fix the issue we have! Just a small comment let me know if you want fix it otherwise we'll take care of it!

PR #523, opened 2025-11-12
Make SAML role mapping optional

gearnode:
## Summary by cubic
Make SAML role mapping optional and add IdP-initiated SAML login support. This prevents unwanted role overrides and lets ACS handle responses without RelayState.

- **New Features**
  - Optional role mapping: MapSAMLRoleToSystemRole returns nil for empty/invalid roles; SAMLUserInfo.Role is now a pointer; EnsureSAMLMembership updates the role only when non-nil and defaults to Member on first creation.
  - IdP-initiated flow: ACS accepts SAMLResponse without RelayState using the c query param (SAML config ID). SP-initiated flow continues to work. Replay protection and validation remain in place.
  - Handler: Logs whether the flow is SP- or IdP-initiated and no longer rejects missing RelayState.

<sup>Written for commit d71800efbdc1ee6d519953befffcfbab7db1fc26. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pkg/server/auth/saml_acs_handler.go
+++ b/pkg/server/auth/saml_acs_handler.go
@@ -64,7 +64,7 @@ func SAMLACSHandler(samlSvc *authsvc.SAMLService, authSvc *authsvc.Service, auth
 		samlConfigID := r.URL.Query().Get("c")
 
 		if relayState != "" {
-			logger.InfoCtx(ctx, "processing SP-initiated SAML login", log.String("relay_state", relayState))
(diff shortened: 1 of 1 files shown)


pkg/authz/service.go
```diff
@@ -842,12 +842,17 @@ func (s *TenantAuthzService) EnsureSAMLMembership(
 					return fmt.Errorf("cannot load membership: %w", err)
 				}
 
+				membershipRole := coredata.RoleMember
```
SachaProbo, line 845: We are removing the member role so we must use either admin or viewer.

PR #579, opened 2025-12-09
Fix audit and framework deletion

SachaProbo:
## Summary by cubic
Makes the audit link on nonconformities optional and safe to delete audits and frameworks without breaking data. The UI now supports selecting “None,” and the backend cleanly handles null audit references.

- **Bug Fixes**
  - Made Nonconformity.audit nullable across DB, GraphQL, service, and MCP.
  - Create/Update now accept null auditId; clearing audit sets auditId to null.
  - Nonconformities list/details handle missing audit and show “No audit.”
  - Audit select adds a “None” option; audit is no longer required in forms.
  - After deleting an audit, redirect back to the audits list.

- **Migration**
  - Run SQL migration to make nonconformities.audit_id nullable and set FK to ON DELETE SET NULL.
  - No data loss; existing links remain. Deleting an audit now sets related nonconformities’ audit_id to null.

<sup>Written for commit 3169fa3c0affceaab660c788064e956ae295c22b. Summary will update automatically on new commits.</sup>


pkg/server/api/mcp/v1/schema.resolvers.go
```diff
@@ -784,7 +784,7 @@ func (r *Resolver) UpdateNonconformityTool(ctx context.Context, req *mcp.CallToo
 			RootCause:          input.RootCause,
 			CorrectiveAction:   UnwrapOmittable(input.CorrectiveAction),
 			OwnerID:            input.OwnerID,
-			AuditID:            input.AuditID,
+			AuditID:            UnwrapOmittable(input.AuditID),
```
gearnode, line 787: @SachaProbo i don't the the mcp.json modification is that expected?

codenem (approved):

PR #598, opened 2025-12-15
Fix ACME renew infinit loop

gearnode:
## Summary by cubic
Stops the infinite ACME renewal loop by shifting renewal to the provisioning flow and capping retries at 3. The renewer now only marks domains as renewing; the provisioner handles the challenge and certificate issuance.

- **Bug Fixes**
  - Removed ACME renew/obtain paths and the HTTP challenge error that caused repeated renew attempts.
  - Renewer sets SSLStatus to Renewing instead of trying to renew in-place, preventing loops.
  - Retry logic capped at 3; domain moves to Failed after exceeding retries.

- **Refactors**
  - Switched DB operations to transactions and use SkipLocked row updates to avoid contention and double-processing.
  - Provisioner now handles both Pending and Renewing statuses, updates status and challenge fields in one place, and maintains the certificate cache.

<sup>Written for commit 3a6295d80840284d6b0dd2d9c1088e579a749d9a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/pkg/certmanager/acme.go
+++ b/pkg/certmanager/acme.go
@@ -21,7 +21,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


gearnode:
Merged

PR #610, opened 2025-12-17
Fix azure storage

SachaProbo:
## Summary by cubic
Fixes Azure Blob Storage compatibility by enabling path-style S3 requests. Adds a new config flag to switch URL style and updates Helm and docs so uploads work reliably on Azure.

- **Bug Fixes**
  - Go: Configure the S3 client with UsePathStyle based on AWS config.
  - Helm: Add s3.usePathStyle value and set AWS_USE_PATH_STYLE in the deployment when true.
  - Docs: Add Azure setup example and note that path-style is required; update values examples.

<sup>Written for commit d6abaec78682c21d0d2a8893461a10c3fc790f1a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/contrib/helm/README.md
+++ b/contrib/helm/README.md
@@ -29,7 +29,7 @@ export PASSWORD_PEPPER=$(openssl rand -base64 32)
 export TRUST_TOKEN_SECRET=$(openssl rand -base64 32)
 ```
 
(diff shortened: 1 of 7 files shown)


gearnode (approved):
LGTM

PR #643, opened 2026-01-05
Fix task display

SachaProbo:
## Summary by cubic
Improved task display to show description, owner, time estimate, and deadline. Simplified assignment by handling it through updateTask with assignedToId.

- **Bug Fixes**
  - Show task description, measure link, owner, time estimate, and deadline in TasksCard.
  - Always show the “Assigned to” field in the task form and send null to unassign.
  - Added formatDuration to render time estimates cleanly.

- **Migration**
  - Removed assignTask and unassignTask mutations; use updateTask with assignedToId.
    - Assign: updateTask { taskId, assignedToId: PeopleID }
    - Unassign: updateTask { taskId, assignedToId: null }

<sup>Written for commit 6ab5ace35386ae24b24eebae89b85d3dc84a1afd. Summary will update on new commits.</sup>

### DIFF
--- a/packages/helpers/src/date.ts
+++ b/packages/helpers/src/date.ts
@@ -26,3 +26,36 @@ function parseDate(dateString: string): Date {
     parts[2] ? parseInt(parts[2], 10) : 1
   );
 }
+
+export function formatDuration(duration?: string | null, __?: (s: string) => string): string | null {
+  if (!duration || !__) return null;
+
+  const timeMatch = duration.match(/PT(\d+)([MH])/);
+  if (timeMatch) {
+    const amount = parseInt(timeMatch[1], 10) || 0;
+    const unit = timeMatch[2];
+    if (unit === "M") {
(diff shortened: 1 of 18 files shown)


packages/helpers/src/date.ts
```diff
@@ -26,3 +26,36 @@ function parseDate(dateString: string): Date {
     parts[2] ? parseInt(parts[2], 10) : 1
   );
 }
+
+export function formatDuration(duration?: string | null, __?: (s: string) => string): string | null {
```
codenem, line 30: Now there are a lot of formatting options in the `Intl` JS API:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat

codenem (commented):
Ok if we do it like this for now for the duration but we should progressively migrate to the Intl and Temporal JS APIs

PR #805, opened 2026-03-09
Fix person update

SachaProbo:
## Summary by cubic
Fixes the person edit flow by excluding read-only fields from update requests and using a shared input for common fields. Create still includes required fields, so both paths work without errors.

- **Bug Fixes**
  - Use a shared input for common fields in `PersonForm`.
  - For updates: send only shared fields + `id` (exclude `emailAddress` and `role`).
  - For creates: include `emailAddress`, `role`, `organizationId`, and `connections`.

<sup>Written for commit 40e71489d8304c02a42ad2628232e8e9d601ffd4. Summary will update on new commits.</sup>


apps/console/src/pages/iam/organizations/people/_components/PersonForm.tsx
codenem, line 169: Let's remove this then if react-hook-form does not scream

PR #834, opened 2026-03-13
Add LLM coding instruction

gearnode:
## Summary by cubic
Adds `AGENTS.md` with coding instructions for LLM agents in the Go codebase: project layout and deps (`chi`, `kit/*`, OpenTelemetry), Go 1.26 pointer creation with `new(expr)`, license/import rules, error/logging patterns, tests, worker design, and run/shutdown orchestration. Adds `CLAUDE.md` as a pointer to `AGENTS.md`.

<sup>Written for commit 7fd4221199afa9a163ad7586f112ed673d18f8bc. Summary will update on new commits.</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,376 @@
+# AGENTS.md
+
+## Project
+
+- Module: `go.probo.inc/probo`
+- Router: `github.com/go-chi/chi/v5`
+- Database: `go.gearno.de/kit/pg` — all raw SQL lives in `pkg/coredata`, never elsewhere
+- HTTP server: `go.gearno.de/kit/httpserver`
+- HTTP client: `go.gearno.de/kit/httpclient`
+- Logging: `go.gearno.de/kit/log`
+- Tracing: OpenTelemetry (`go.opentelemetry.io/otel`)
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,373 @@
...
+
+var (
+	_ Reader = (*FileReader)(nil)
+	_ Writer = (*FileWriter)(nil)
+)
+```
+
+### One argument per line
+
+When a function call has multiple arguments, put each argument on its own line. Never pack arguments on the opening line.
```
SachaProbo: When a function call has multiple argument on more than one line. 

This would be ok:
```go
foo.NewService(ctx, db, logger)
```

AGENTS.md
```diff
@@ -0,0 +1,373 @@
...
+
+// Bad
+svc, err := foo.NewService(ctx, db, logger, foo.Config{
+	Interval: 10 * time.Second,
+})
+```
+
+### License header
+
+Every `.go` file starts with the ISC license header. Use the **current year** for new files. When editing an existing file whose header has a different year, update it to a range (e.g. `2025-2026`). Never overwrite the original year.
```
SachaProbo, line 75: Do we need AI for this ? We could use a script.

PR #881, opened 2026-03-18
Fix race condition in magic link token verification

gearnode:
## Summary

- Fixed race condition in `OpenSessionWithMagicLink` where token lookup used a separate connection instead of the transaction, breaking the `FOR UPDATE` lock that prevents concurrent magic link verification
- Corrected "resouce" → "resource" typo in authentication error messages

## Changes

The `SELECT ... FOR UPDATE` lock must be held within the transaction to provide mutual exclusion. Now uses the transaction connection directly for the token lookup, ensuring the lock is held until commit/rollback.

---
## Summary by cubic
Fixes a race condition in magic link token verification by loading the token with SELECT FOR UPDATE using the same transaction (`tx`), ensuring the lock is held until commit. Also fixes an authentication error message typo: "resouce" → "resource".

<sup>Written for commit bf6965aff14943d4f7f7f2a1aa68acd781203703. Summary will update on new commits.</sup>

### DIFF
--- a/pkg/iam/auth_service.go
+++ b/pkg/iam/auth_service.go
@@ -628,21 +628,12 @@ func (s AuthService) OpenSessionWithMagicLink(ctx context.Context, tokenString s
 			hashedValue := HashToken(tokenString)
 			token := &coredata.Token{}
 
-			if err := s.pg.WithConn(
-				ctx,
-				func(conn pg.Conn) error {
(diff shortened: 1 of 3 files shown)


SachaProbo (commented):
LGTM

codenem (approved):

PR #1000, opened 2026-04-07
Consolidate document draft management and add delete draft

SachaProbo:
Merge createDraftDocumentVersion, updateDocumentVersion, and deleteDraftDocumentVersion into the existing updateDocument mutation. The backend auto-creates a draft when title, content, classification, or type changes, updates an existing draft, and auto-deletes it when all fields revert to match the published version.

Add a new deleteDocumentDraft mutation (GraphQL, MCP) that takes a document ID and deletes the latest version if it is a draft (excluding the initial v0.1). Exposed in the actions dropdown before archive.

---
## Summary by cubic
Consolidates draft lifecycle into `updateDocument` and adds `deleteDocumentDraft` by document ID. Auto-save keeps editors stable across draft create/delete, and the latest published content is shown after draft removal.

- **New Features**
  - `updateDocument` now accepts `title`, `content`, `classification`, and `documentType`. It auto-creates/updates a draft on change and auto-deletes it when values match the published version. The payload includes `documentVersion` only when a draft exists or is created/updated.
  - Editors (title, description, classification, type) call `updateDocument`. The description editor is enabled only on the latest version when the document isn’t archived and you have `core:document:update`; historical versions are read-only.
  - Loaders use store-and-network and refetch on version transitions; a version-change signal from layout keeps the description view fresh. A stable editor key uses a data-generation counter to prevent remounts during auto-save. Cancel resets fields, and forms stay in sync via `react-hook-form` values.
  - Actions dropdown removes “Create draft”. It shows “Delete draft” only when the latest is a deletable draft (not v0.1) and you have `core:document:delete-draft`. It calls `deleteDocumentDraft`, refetches, and navigates to the description on success.
  - MCP: `updateDocument` accepts markdown `content` and converts it to ProseMirror automatically.

- **Migration**
  - Remove `createDraftDocumentVersion`, `updateDocumentVersion`, and `deleteDraftDocumentVersion` (GraphQL and MCP). Use `updateDocument` for title/content/classification/type and `deleteDocumentDraft` with `documentId`. Permissions now use `core:document:update` and `core:document:delete-draft` (version-level update/delete-draft actions are removed).
  - CLI: remove `probo document create-draft` and `probo document update-version`. Use `probo document update --title/--content/--document-type/--classification` and `probo document delete-draft <document-id>`.

<sup>Written for commit 03708d45c321191a7d4f266e2364537e3953dc93. Summary will update on new commits.</sup>


e2e/console/document_version_test.go
```diff
@@ -447,6 +521,128 @@ func TestDocumentVersion_BulkRequestSignatures(t *testing.T) {
...
+					}
+					documentVersion {
+						id
+						status
+					}
+				}
+			}
+		`
+
+		var result struct {
```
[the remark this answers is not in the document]
SachaProbo: The result variables are all local to each subtest. The query string is read-only.

pkg/probo/document_service.go
```diff
@@ -1625,21 +1551,112 @@ func (s *DocumentService) Update(
...
+					return err
+				}
 
-				if err := draftVersion.Update(ctx, tx, s.svc.scope); err != nil {
-					return fmt.Errorf("cannot update draft version: %w", err)
+				if err := s.updateVersionInTx(ctx, tx, draftVersion, req.Content, req.Classification, req.DocumentType, document.Title); err != nil {
+					return err
 				}
+
+				resultVersion = draftVersion
```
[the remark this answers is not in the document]
SachaProbo: The issue is not valid. When we reach the “no draft exists” branch (line 1600), the latest version is published. createDraftInTx clones the published version, then updateVersionInTx applies the request fields. Since we already checked hasVersionChanges is true (line 1560-1564), at least one field differs from what was copied. The draft will always differ from the published version after the update

apps/console/src/pages/organizations/documents/description/DocumentDescriptionPage.tsx
```diff
@@ -102,6 +116,13 @@ export function DocumentDescriptionPage(props: { queryRef: PreloadedQuery<Docume
             return;
           }
 
+          // Refetch layout when draft status changes (created or deleted)
+          // so the drawer and header reflect the current version.
```
codenem: What happens if I start writing again during this call, do I lose this content after the refetch?

apps/console/src/pages/organizations/documents/description/DocumentDescriptionPage.tsx
```diff
@@ -102,6 +116,13 @@ export function DocumentDescriptionPage(props: { queryRef: PreloadedQuery<Docume
...
+          const draftReturned = !!data.updateDocument.documentVersion;
+          if (wasDraft !== draftReturned) {
+            onRefetch();
```
codenem, line 128: Why should we refetch instead of just updating the store?
If the mutation sends back needed data then the store should update itself without the need for refetch.

apps/console/src/pages/organizations/documents/description/DocumentDescriptionPageLoader.tsx
```diff
@@ -40,6 +41,15 @@ function DocumentDescriptionPageQueryLoader() {
     }
   });
 
+  useEffect(() => {
+    if (draftDeletedAt > 0) {
```
codenem: I don't understand the use of this? Why do we need to refetch here, when a draft is deleted then we should navigate to the last available version if we were on the draft page. This should be enough without refetch

SachaProbo: The version is not in the url.

apps/console/src/pages/organizations/documents/description/DocumentDescriptionPage.tsx
SachaProbo: The refetch is there only for the edge case where the connection membership changes (draft created or deleted), not on every save. In those situation the page refetch and it is not really possible to writes during the refetch.

PR #1021, opened 2026-04-10
fix: set cookie secure flag to false in dev config

emimoir:
The dev config (`cfg/dev.yaml`) uses `http://localhost:8080` as its base URL but both session cookies have `secure: true`. Browsers refuse to store Secure cookies over plain HTTP, which causes a sign-in loop after successful authentication.

This sets both `cookie.secure` values to `false` so the dev setup works out of the box.

---
## Summary by cubic
Fixes the local sign-in loop by setting session cookies to `secure: false` in dev, and adds the Pebble dev root CA so the ACME client can trust the local directory.

- **Bug Fixes**
  - Updated `cfg/dev.yaml`: set both session cookies to `secure: false` for `http://localhost:8080`, and added `acme.root-ca` for `https://localhost:14000/dir`.

<sup>Written for commit 86c92d473c7d64fbadc561f51a89a8d78acf7f34. Summary will update on new commits.</sup>

### DIFF
--- a/cfg/dev.yaml
+++ b/cfg/dev.yaml
@@ -26,7 +26,7 @@ probod:
         domain: "localhost"
         secret: "this-is-a-secure-secret-for-cookie-signing-at-least-32-bytes"
         duration: 24
-        secure: true
+        secure: false
 
   trust-center:
     http-addr: ":8085"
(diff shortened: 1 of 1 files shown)


cfg/dev.yaml
```diff
@@ -87,6 +109,32 @@ probod:
       directory: "https://localhost:14000/dir"
       email: "admin@getprobo.com"
       key-type: "EC256"
+      root-ca: |
+        -----BEGIN CERTIFICATE-----
```
SachaProbo, line 113: You should not commit the cert.

PR #1034, opened 2026-04-13
Add membership and host to user webhook payload

SachaProbo:
Nest membership ID, role, and state into a membership sub-object in the user webhook payload. Also emit user:updated webhook when the membership role is changed.

---
## Summary by cubic
Add nested membership info to the user webhook payload and send the `X-Probo-Webhook-Host` header. Also emit `user:updated` when a membership role changes.

- **New Features**
  - Payload includes `membership` with `id`, `role`, and `state` (`user.membership`).
  - `user:updated` is emitted on membership role changes.
  - All user webhooks (created/updated/deleted) include `membership` when available; on deletes it may be null.
  - Webhook requests include `X-Probo-Webhook-Host`.

- **Migration**
  - `user.state` removed; use `user.membership.state`.
  - Read membership fields from `user.membership.id` and `user.membership.role`.
  - Expect additional `user:updated` events on role changes; adjust filters/handlers if needed.

<sup>Written for commit 06c0972551d062b9eb5d160db221e7e703d59318. Summary will update on new commits.</sup>

### DIFF
--- a/pkg/iam/organization_service.go
+++ b/pkg/iam/organization_service.go
@@ -273,6 +273,15 @@ func (s *OrganizationService) UpdateMempership(
 				return fmt.Errorf("cannot update membership: %w", err)
 			}
 
+			profile := &coredata.MembershipProfile{}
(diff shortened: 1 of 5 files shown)


gearnode (approved):
LGTM

PR #1053, opened 2026-04-16
Preserve shared connector on SCIM disconnect

aureliensibiril:
## Summary

`deleteSCIMConfiguration` unconditionally deleted the underlying OAuth2 connector together with the SCIM bridge and config. When the same Google Workspace connector was also referenced from `access_sources` (the case when Google Workspace is configured for both SCIM and Access Reviews), the `access_sources.connector_id` foreign key rejected the DELETE, aborting the whole transaction. Nothing was deleted and the resolver returned an INTERNAL error — but the frontend `updater` ran regardless and optimistically cleared `organization.scimConfiguration`, showing the user a false "disconnected" UI until the next refresh revealed the state had not changed.

## Changes

- **`feat(coredata)`** add `AccessSources.CountByConnectorID` — helper used to gate the disconnect flow on remaining references.
- **`fix(iam)`** `DeleteSCIMConfiguration` now skips the connector delete when any access source still references it. The bridge's own `connector_id` FK is `ON DELETE SET NULL`, so dropping the bridge alone is sufficient to unbind SCIM.
- **`fix(console)`** replace the custom `updater` on the Google Workspace disconnect mutation with a `@deleteRecord` directive on `deletedScimConfigurationId` (matching the idiom already used in `SCIMConfiguration.tsx`). `@deleteRecord` is a no-op when the returned id is null, so error responses no longer flip the UI.

## Test plan

- [x] Reproduce the original bug locally: click Disconnect on an org that has both a Google Workspace SCIM bridge and a Google Workspace access source referencing the same connector → mutation returns `INTERNAL`, DB rows untouched, UI shows false "disconnected" state until refresh restores "Connected".
- [x] `go vet ./pkg/coredata/... ./pkg/iam/...` → clean.
- [x] `make test MODULE=./pkg/coredata/...` → 70 tests pass.
- [x] `make test MODULE=./pkg/iam/...` → 111 tests pass.
- [x] `npm run relay` → regenerates `GoogleWorkspaceConnectorDeleteMutation.graphql.ts` with `"handle": "deleteRecord"`.
- [x] `cd apps/console && npx eslint` on the modified file → 0 errors; `npm run build` → no type errors.
- [x] End-to-end smoke in the dev environment after the fix:
[...]

PR #1054, opened 2026-04-16
Use typed error when SCIM bridge is not found

aureliensibiril:
## Summary

`OrganizationService.UpdateSCIMBridge` returned bare `fmt.Errorf("SCIM bridge not found")` strings on both not-found branches (resource missing in DB, and cross-tenant mismatch). Every other call site returning the same condition uses the typed `NewSCIMBridgeNotFoundError(bridgeID)` — see `pkg/iam/organization_service.go:1962` and `:2153`.

Switch both branches to the typed error so:
- Error shape is consistent across the service.
- Callers can detect the condition with `errors.As` instead of string matching.
- The GraphQL/CLI error mapping paths that already handle `*ErrSCIMBridgeNotFound` work uniformly for this resolver too.

## Test plan

- [x] `go build ./pkg/iam/...` → clean.
- [x] `go vet ./pkg/iam/...` → clean.
- [x] No behavior change for callers that already swallow the error into `gqlutils.Internal` — the typed error still unwraps to a human-readable string.

## Notes

Spotted during review of a separate SCIM disconnect bugfix (#1053). Split into its own PR because it's a preexisting inconsistency, not caused by that change.

---
## Summary by cubic
Return the typed `NewSCIMBridgeNotFoundError(bridgeID)` from `OrganizationService.UpdateSCIMBridge` in both not-found cases (bridge missing or cross-tenant mismatch). This keeps errors consistent with other call sites and lets callers and GraphQL/CLI mappers use `errors.As` to detect it.

<sup>Written for commit f5c59ba13bfba957325d8002926621f83d2acb3c. Summary will update on new commits.</sup>

### DIFF
--- a/pkg/iam/organization_service.go
+++ b/pkg/iam/organization_service.go
@@ -1670,14 +1670,14 @@ func (s OrganizationService) UpdateSCIMBridge(
 			err := bridge.LoadByID(ctx, tx, scope, bridgeID)
 			if err != nil {
 				if err == coredata.ErrResourceNotFound {
-					return fmt.Errorf("SCIM bridge not found")
+					return NewSCIMBridgeNotFoundError(bridgeID)
 				}
 
 				return fmt.Errorf("cannot load SCIM bridge: %w", err)
 			}
 
 			if bridge.OrganizationID != organizationID {
-				return fmt.Errorf("SCIM bridge not found")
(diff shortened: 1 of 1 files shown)


SachaProbo (approved):

SachaProbo:
Merged with 0a5dfc652f42782bab91796110390c038bca7cab

PR #1091, opened 2026-04-22
Handle SCIM user email rename via external ID fallback

SachaProbo:
When a user's email is renamed in the identity provider (e.g. Google Workspace), the external ID stays the same but the email changes. The SCIM CreateUser now falls back to external ID lookup when no profile is found by identity, and reassociates the existing profile to the new identity instead of failing with a 409 uniqueness error.

Also removes user emails from bridge sync error messages to avoid logging PII, using external IDs instead.

---
## Summary by cubic
Fixes SCIM user provisioning when an IdP renames a user’s email by falling back to external ID, reusing the existing profile, migrating the membership to the new identity, and updating the email to avoid 409s. Also replaces emails with external IDs in bridge sync errors to reduce PII.

- **Bug Fixes**
  - CreateUser: if no profile by identity, look up by external ID; if found, rebind the profile to the new identity, migrate the membership (preserve role), keep the same profile ID, and update email.
  - Data layer: added profile lookup by external ID; allowed updating `identity_id` in membership and profile updates.
  - Tests/bridge: added e2e tests for create/list/get/delete, email rename flow, duplicate handling, and auth; bridge sync errors now reference external IDs (create/update/delete/deactivate).

<sup>Written for commit e5e17d59ac6d28ccb238883cfa18bac68487248f. Summary will update on new commits.</sup>

### DIFF
--- a/pkg/iam/scim/service.go
+++ b/pkg/iam/scim/service.go
@@ -177,6 +177,7 @@ func (s *Service) CreateUser(
 
 		eventType := coredata.WebhookEventTypeUserUpdated
 		profile = &coredata.MembershipProfile{}
+
 		if err := profile.LoadByIdentityIDAndOrganizationID(
 			ctx,
 			tx,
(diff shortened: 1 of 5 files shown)


pkg/iam/scim/service.go
```diff
@@ -177,14 +177,30 @@ func (s *Service) CreateUser(
...
+		if profileErr != nil && errors.Is(profileErr, coredata.ErrResourceNotFound) && externalIdPtr != nil {
+			profileErr = profile.LoadByExternalIDAndOrganizationID(
+				ctx,
+				tx,
+				scope,
+				*externalIdPtr,
+				config.OrganizationID,
+			)
+		}
+
```
gearnode: inline the if

SachaProbo:
@cubic-dev-ai please review

PR #1097, opened 2026-04-23
Sanitize MCP errors to avoid leaking internal details

SachaProbo:
## Summary by cubic
Sanitize MCP errors so tool responses never leak internal details. Use the `go.probo.inc/mcpgen` recover hook to map known errors to clear messages; unknown errors return "internal server error".

- **Bug Fixes**
  - Centralized sanitization in `pkg/server/api/mcp/mcputils/recovery.go`: map "permission denied", "assumption required", "resource not found", "resource already exists", and "resource is in use"; return validation errors as-is; log unknown errors and return "internal server error".
  - Replaced `RecoveryMiddleware` with the generated recover hook: `server.New(..., mcpgenmcp.WithRecoverFunc(mcputils.NewRecoverFunc(logger)))`.
  - Added `CallToolExpectToolError` in `e2e/internal/testutil/mcp.go`.
  - E2E for assets, risks, and vendors assert sanitized "resource not found" after delete/update, "permission denied" for viewer creates, and validation messages without "pq:" or "sql:".

- **Dependencies**
  - Bumped `go.probo.inc/mcpgen` to `v0.0.0-20260428172408-1496ba9b4619`.

<sup>Written for commit d82df6b8ee61db35f316b0630d1718513dbd1aab. Summary will update on new commits.</sup>

### DIFF
--- a/e2e/internal/testutil/mcp.go
+++ b/e2e/internal/testutil/mcp.go
@@ -204,6 +204,20 @@ func (mc *MCPClient) CallTool(toolName string, args map[string]any) *MCPToolResu
 	return &toolResult
 }
 
+// CallToolExpectToolError invokes an MCP tool and expects a tool-level error
+// (isError: true in the result). It returns the error text content.
+func (mc *MCPClient) CallToolExpectToolError(toolName string, args map[string]any) string {
+	tr := mc.CallTool(toolName, args)
+	require.True(mc.t, tr.IsError, "expected tool %s to return isError", toolName)
+	require.NotEmpty(mc.t, tr.Content, "tool %s returned no content", toolName)
+
+	var text string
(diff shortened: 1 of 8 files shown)


codenem (commented):
We should have errors for mcp surface has we do for graphql in gqlutils errors.go or jsonutil errors.go no? Or maybe the MCP standard does not care?

gearnode:
@codenem Agree with you. I think @SachaProbo patches are more of a quick win to at least avoid leaking internal errors and to have a panic recovery mechanism in place. It feels like a good first step before moving toward proper error handling, like what we already do in Connect and in the new console resolvers.

gearnode (approved):
LGTM

PR #1023, opened 2026-04-10
Add agent graceful stop/resume with checkpoint persistence

aureliensibiril:
## Summary

- LLM `Message` and all `Part` types now round-trip through JSON for checkpoint serialization
- `coreLoop` saves incremental checkpoints after every tool-call turn and checks a cooperative stop signal at turn boundaries
- `Restore` function reconstructs agent state from a persisted checkpoint (suspended, nested suspended, awaiting approval)
- `AgentRun` entity and `PGCheckpointStore` in coredata with lease-based concurrency and 10 MiB size guard
- Poll-based supervisor worker that claims, runs, heartbeats, and recovers agent runs; on graceful shutdown rows stay RUNNING so stale recovery auto-resumes on restart

## Test plan

- [ ] `make test MODULE=./pkg/llm` — JSON round-trip tests (40 tests)
- [ ] `make test MODULE=./pkg/agent` — stop signal, restore, existing agent tests (284 tests)
- [ ] Verify migration applies cleanly: `make stack-up && make psql` then run `pkg/coredata/migrations/20260410T120000Z.sql`
- [ ] Manual: start an agent run, kill the process mid-turn, restart, verify it resumes from checkpoint

---
## Summary by cubic
Adds graceful stop/resume to agents with durable JSON checkpoints, a Postgres-backed `AgentRun` store, and a `go.gearno.de/kit/worker` supervisor that auto-resumes runs on restart. Snapshots `MaxTurns` into checkpoints and reapplies it on restore; failures surface to worker metrics/spans.

- **New Features**
  - Checkpoints and JSON: LLM `Message`/`Part` (`Text`/`Image`/`File`/`Thinking`), `ToolCall`, `FunctionCall`, and `Usage` round-trip via JSON (type-tagged `Part` envelope); new `Checkpoint`/`Checkpointer`; export `CompletedCall`; `Result` JSON omits `LastAgent` and uses stable lowercase field names.
  - Agent loop and resume: per-turn checkpoints; cooperative stop via `WithStopSignal`; `Restore` handles nested and awaiting-approval states; adds `StreamEventSuspended`, `OnRunRestore`, and `OnRunSnapshot`; guards nil checkpoints in parallel suspension and fixes nested/parallel suspension fallback; in-memory checkpoint fallback on save failure; snapshots `MaxTurns` and reapplies it on restore.
  - APIs: `Run`, `RunStreamed`, and `Resume` now take variadic `RunOption`; use `WithCheckpointer` to enable persistence.
[...]

PR #1110, opened 2026-04-27
Drive graceful agent suspend from ctx cancellation

aureliensibiril:
## Summary
- Collapse the dual stop-mechanism into a single signal: `ctx.Done()` now means **graceful suspend** in `pkg/agent`. `coreLoop` shadows the incoming ctx with `context.WithoutCancel` so an in-flight LLM call, tool, and checkpoint write all complete normally; only the at-boundary check observes the original ctx.
- Delete the dedicated stop-signal API (`WithStopSignal` / `stopSignalFrom`) and the supervisor-side `stopCh` plumbing. The per-run forwarder now calls `cancelRun(ErrSuspendForCheckpoint)` when `h.shutdownCh` closes.
- Document the new contract on `Run`/`RunStreamed`/`Resume`/`Restore` and in `contrib/claude/agent.md`.

## Why
Reviewer feedback on #1023: ctx already encodes a cancel signal — having a second channel for "graceful stop" is duplicate plumbing. This PR makes ctx the single source of truth. Side-effects: deadlines (`context.WithTimeout`) become "max budget then suspend with checkpoint" instead of killing work outright; there is no longer an in-process hard-abort path (callers that need to kill a run terminate the process).

## Test plan
- [x] `go test ./pkg/agent/... ./pkg/coredata/... ./pkg/probo/... ./pkg/agentruntest/...`
- [x] `go test -race ./pkg/agent/... ./pkg/agentruntest/...`
- [x] `go vet ./...`, `gofmt -l pkg/`
- [x] New `pkg/agent/cancel_test.go` covers (a) pre-first-turn cancel and (b) mid-run cancel from inside a tool
- [x] Existing SIGTERM/parallel-suspend integration tests in `pkg/agentruntest` pass unchanged

---
## Summary by cubic
Make `ctx` the single graceful-suspend signal across the agent loop. Cancelling `ctx` now checkpoints at the next safe boundary and returns `*SuspendedError`; in‑flight LLM calls, tools, and saves run on `context.WithoutCancel` so they finish cleanly.

- **Refactors**
  - Removed `WithStopSignal`/`stopSignalFrom` and related tests in `pkg/agent`.
  - Updated `Run`, `RunStreamed`, `Resume`, and `Restore` to split signal/work contexts via a suspend shield; only boundary checks observe the original `ctx`.
  - Added `ErrSuspendForCheckpoint` in `pkg/agent`; supervisor maps shutdown to `cancelRun(ErrSuspendForCheckpoint)` in `pkg/probo/agent_run_handler.go`.
  - Made nested restore progress writes use `context.WithoutCancel`.
  - Documented cancellation semantics in `contrib/claude/agent.md`; tests cover pre-first-turn, mid-run, and in‑flight LLM cancel, and assert input messages land in the suspension checkpoint.

- **Migration**
  - To gracefully suspend, cancel the run’s `ctx`; remove any use of `WithStopSignal`.
  - Deadlines via `context.WithTimeout` now suspend after the budget instead of hard-aborting.
  - There is no in-process hard-abort; terminate the process to kill a run.
  - Tools needing their own deadline should derive one inside the tool (`context.WithTimeout`).

<sup>Written for commit 0f1d893668026a8489b1757fe9c2af730aacda3d. Summary will update on new commits.</sup>


pkg/agent/context.go
```diff
@@ -51,3 +51,10 @@ func TryRunContextFrom[C any](ctx context.Context) (C, bool) {
 	typed, ok := val.(C)
 	return typed, ok
 }
+
+// suspendShield splits ctx in two: signal keeps the cancel signal for
+// turn-boundary checks, work survives cancellation and carries every
+// downstream call (LLM, tools, hooks, save) through to completion.
+func suspendShield(ctx context.Context) (signal, work context.Context) {
```
gearnode: not sure to understand why creating a function for that tbh.

aureliensibiril: purely to document the behavior in only one place instead of having to add comments on the multiple call sites of withoutCancel.

gearnode: let do not hide, this in a function, I prefer keep that explicit.

pkg/agent/run.go
```diff
@@ -368,34 +373,26 @@ func coreLoop(ctx context.Context, startAgent *Agent, inputMessages []llm.Messag
...
-
-				if s.opts.checkpointer != nil && s.opts.runID != "" {
-					if saveErr := s.opts.checkpointer.Save(ctx, s.opts.runID, cp); saveErr != nil {
-						s.logger.ErrorCtx(ctx, "cannot save suspension checkpoint", log.Error(saveErr))
-						se.Checkpoint = cp
-					} else {
-						emitHook(s.agent, func(h RunHooks) { h.OnRunSnapshot(ctx, s.agent, cp) })
-					}
-				} else {
+			if s.opts.checkpointer != nil && s.opts.runID != "" {
```
gearnode: We have case where we have runID but no checkpointer? Even do we have case without checkpointer?

pkg/agent/run.go
```diff
@@ -368,34 +373,26 @@ func coreLoop(ctx context.Context, startAgent *Agent, inputMessages []llm.Messag
...
 
 	for {
-		if err := ctx.Err(); err != nil {
-			return s.finishRun(ctx, nil, fmt.Errorf("cannot complete: %w", err))
-		}
-
-		if s.turns >= s.agent.maxTurns {
-			return s.finishRun(ctx, nil, &MaxTurnsExceededError{MaxTurns: s.agent.maxTurns})
-		}
+		if err := outerCtx.Err(); err != nil {
```
gearnode: select ctx.Done()

pkg/probo/agent_run_handler.go
```diff
@@ -101,13 +102,12 @@ func (h *agentRunHandler) Process(ctx context.Context, run coredata.AgentRun) er
 	runCtx, cancelRun := context.WithCancelCause(ctx)
 	defer cancelRun(nil)
 
-	stopCh := make(chan struct{})
 	forwarderDone := make(chan struct{})
 	defer close(forwarderDone)
 	go func() {
 		select {
 		case <-h.shutdownCh:
-			close(stopCh)
```
gearnode, line 110: go routine leak?

aureliensibiril: No leak I think. The forwarder goroutine has two exit paths:

- `<-h.shutdownCh` — only fires if the handler is shutting down
- `<-forwarderDone` — closed by `defer close(forwarderDone)` on every `Process` return path

Lifetimes:
- normal completion: `executeRun` returns → defers fire → `forwarderDone` closes → goroutine exits via the second case
- shutdown: `shutdownCh` closes → `cancelRun(ErrSuspendForCheckpoint)` fires → `executeRun` observes ctx cancel → returns → defers fire (the goroutine already exited, `close` of `forwarderDone` is just a no-op observer)

Either way the goroutine is bounded by `Process`'s scope.

gearnode (approved):
LGTM

PR #1126, opened 2026-04-29
Return clear error on user delete FK violation

SachaProbo:
Deleting a membership profile referenced by other tables (owner, approver, assignee, etc.) surfaced as a generic Internal error. Add NewResourceInUseError to wrap the sentinel with the blocking table name, return it on Postgres FK violation (23503) in the coredata Delete, and map it to a CONFLICT in the GraphQL and MCP resolvers.

---
## Summary by cubic
Return a clear error when deleting a user profile that’s still referenced by other records. Postgres FK violations now map to a Conflict in GraphQL and show “cannot remove user” in MCP.

- **Bug Fixes**
  - In `pkg/coredata/membership_profile.go`, detect `pgconn.PgError` code `23503` and return `coredata.ErrResourceInUse`.
  - In `pkg/server/api/connect/v1`, map `ErrResourceInUse` to Conflict; in `pkg/server/api/mcp/v1`, surface “cannot remove user”.

<sup>Written for commit 9e61e5e1b737019e444b141c13fa7cd49b2be227. Summary will update on new commits. <a href="https://cubic.dev/pr/getprobo/probo/pull/1126?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/pkg/coredata/membership_profile.go
+++ b/pkg/coredata/membership_profile.go
@@ -1416,6 +1416,11 @@ WHERE
 
 	_, err := conn.Exec(ctx, q, args)
 	if err != nil {
+		if pgErr, ok := errors.AsType[*pgconn.PgError](err); ok {
+			if pgErr.Code == "23503" {
+				return ErrResourceInUse
+			}
+		}
 		return fmt.Errorf("cannot delete profile: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


gearnode:
Not bug fan of leaking those information in the error. We should isntead address why we got this error and handle it.

SachaProbo:
> Not bug fan of leaking those information in the error. We should isntead address why we got this error and handle it.

The only way to allow deletion would be to remove the restrice and change it by a nullify or cascade, but I am not sure this is what we want.

The other solution would be to disable user deletion entirely.

PR #1146, opened 2026-05-04
Defensive / migration guard on uncategorised cookies

codenem:
## Summary by cubic
Exclude the UNCATEGORISED category at the SQL level and rename GraphQL `categories` to `consentCategories`. Add a Detection tab to review uncategorised patterns with actions, and move cookie management into the Display page.

- **Bug Fixes**
  - SQL-level UNCATEGORISED filter; rename to `consentCategories`; update services/pages, remove client-side filters, add a defensive guard, adjust the snapshot rank test, fix e2e tests for the rename and UNCATEGORISED exclusion, and remove a redundant check.
  - Guard the client store updater when moving patterns and surface GraphQL errors on reorder mutations.

- **New Features**
  - Track `last_detected_at` on cookies and `last_matched_at` on `cookie_patterns` (cookie timestamp backfilled; pattern timestamp refreshed during analysis). Expose `lastMatchedAt` in GraphQL, MCP, CLI list/view, and `n8n` get/getAll.
  - Add `uncategorisedPatterns` on `CookieBanner` and a new Console Detection page with pagination, sorting by NAME/LAST_MATCHED_AT/UPDATED_AT/SOURCE, text/source filters, and row actions (edit, move to category, exclude, delete). Consolidate cookie CRUD and reordering into the Display page; remove the Cookies tab. The move-to dropdown uses an interaction-triggered preloaded query; documented in the React components guide.

<sup>Written for commit f83b8ee3cabec86f26e008b52d653c76ee71543d. Summary will update on new commits.</sup>


pkg/coredata/cookie_category.go
```diff
@@ -323,6 +412,61 @@ ORDER BY
...
+	AND kind != @excluded_kind
+ORDER BY
+	rank ASC, id ASC;
+`
+
+	q = fmt.Sprintf(q, scope.SQLFragment())
+
+	args := pgx.StrictNamedArgs{
+		"cookie_banner_id": cookieBannerID,
+		"excluded_kind":    CookieCategoryKindUncategorised,
```
SachaProbo, line 451: No need for an argument if the kind is always the same

pkg/coredata/cookie_pattern.go
```diff
@@ -52,6 +53,17 @@ func (cp *CookiePattern) CursorKey(field CookiePatternOrderField) page.CursorKey
 	switch field {
 	case CookiePatternOrderFieldCreatedAt:
 		return page.NewCursorKey(cp.ID, cp.CreatedAt)
+	case CookiePatternOrderFieldName:
+		return page.NewCursorKey(cp.ID, cp.DisplayName)
+	case CookiePatternOrderFieldLastMatchedAt:
+		if cp.LastMatchedAt == nil {
```
SachaProbo, line 59: Do we really need this ?

SachaProbo (approved):

pkg/coredata/cookie_category.go
codenem: Yes it's a rule in coredata.md so we don't hardcode values of enums

pkg/coredata/cookie_pattern.go
codenem: The if block or the column? The block matches the COALESCE value for the order field.
The column itself will be useful to help the admin / the future categorizing agent to make choices. I'll also add a counter.
Because some cookies that will be detected can come from old integrations that set a cookie on a visitor browser but aren't used anymore.

SachaProbo: I was talking about the if block.

PR #1223, opened 2026-05-26
Fix common third-party logo URL not-found error

SachaProbo:
CommonThirdParty.AuthorizationAttributes returned an empty map, but the IAM authorizer requires an entry for every requested resource ID and treats missing entries as ErrResourceNotFound. With the new authorize() guard on commonThirdPartyResolver.LogoURL, this surfaced as "commonThirdParties[*].logoUrl resource not found" from the CommonThirdPartyComboboxQuery.

Switch AuthorizationAttributes to a SELECT-by-id existence check that returns an empty attributes map for each row that exists. The catalog is global with no organization, so omitting organization_id is the intended shape: the authorizer then loads no membership, evaluates only identity-scoped policies (CommonThirdPartyCatalogPolicy grants access), and skips audit logging.

Add an e2e regression test that seeds a CommonThirdParty directly via the test database and asserts the GraphQL logoUrl path resolves without error.

---
## Summary by cubic
Fixes the "resource not found" error on `commonThirdParties[*].logoUrl` by loading existence-only authorization attributes per ID in the global catalog. Adds an e2e regression test to ensure the GraphQL path resolves without error.

- **Bug Fixes**
  - Implement SELECT-by-id in `CommonThirdParty.AuthorizationAttributes` and return empty attribute maps for existing rows (no `organization_id`), preventing `ErrResourceNotFound` from the IAM authorizer on `logoUrl`.
  - Add e2e test that seeds a `CommonThirdParty` and verifies `commonThirdParties[*].logoUrl` resolves successfully.

<sup>Written for commit e6f7884a2980d22f859a6ba23c858b7bbd9bf21e. Summary will update on new commits. <a href="https://cubic.dev/pr/getprobo/probo/pull/1223?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/e2e/console/common_third_party_test.go
+++ b/e2e/console/common_third_party_test.go
@@ -1,0 +1,113 @@
+// Copyright (c) 2026 Probo Inc <hello@getprobo.com>.
+//
+// Permission to use, copy, modify, and/or distribute this software for any
+// purpose with or without fee is hereby granted, provided that the above
+// copyright notice and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+// REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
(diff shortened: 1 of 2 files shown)


SachaProbo:
@cubic-dev-ai please review

codenem (approved):

PR #1236, opened 2026-05-27
refactor(awsconfig): replace manual credential chain with LoadDefault…

neilmcgibbon:
## Summary

Replace the hand-rolled AWS credential chain in `pkg/awsconfig` with a single call to `config.LoadDefaultConfig` from the AWS SDK v2.

### What was wrong

`NewConfig` was manually wiring together an ECS endpoint provider and an EC2 IMDS provider, reproducing (badly) what the SDK already does. It missed EKS Pod Identity, IRSA (Web Identity tokens), `~/.aws` profiles IAM Identity Center/SSO, and `AWS_PROFILE`. It also contained custom endpoint-validation logic to guard against credential exfiltration — code
that exists precisely because we were bypassing the SDK's own safety checks.

### What changed

- `config.LoadDefaultConfig` now drives credential resolution. The full  SDK chain is active: env vars → `~/.aws` → container credentials (ECS/EKS Pod Identity) → EC2 instance profile → SSO.
- Static credentials (`AccessKeyID` + `SecretAccessKey`) still take priority via `WithCredentialsProvider`.
- The SSRF-protected HTTP client is preserved via `WithHTTPClient`.
- `NewConfig` now returns `(aws.Config, error)` to surface load failures.
- All the hand-rolled endpoint-validation and credential-chain code is deleted (~57 lines removed, 32 added).

### Why this is safe

- In production containers `~/.aws` doesn't exist, behaviour is  identical to before for ECS/EKS/EC2 environments.
- Static-credential deployments are unaffected.
- Developer machines with a `[default]` AWS profile will now use those credentials when no static keys are configured — which is the standard SDK behaviour and an improvement over silently failing.

---
## Summary by cubic
Replaced the manual AWS credential chain in `pkg/awsconfig` with `config.LoadDefaultConfig`, enabling the full SDK chain (env, `~/.aws`, ECS/EKS Pod Identity, EC2, SSO) while keeping our SSRF-safe HTTP client. Updated the caller to handle the new error return and adjusted dependencies.

### Service **`+24`** **`-52`**
- Use `config.LoadDefaultConfig` with `WithRegion` and `WithHTTPClient`.
- Keep static creds via `credentials.NewStaticCredentialsProvider` (takes priority).
- Preserve endpoint override by setting `BaseEndpoint` when provided.
- Change `NewConfig` to return `(aws.Config, error)` and handle the error in `pkg/probod`.

### Other **`+8`** **`-5`**
- Add `github.com/aws/aws-sdk-go-v2/config`; bring in SSO/STS deps indirectly.
- Move `feature/ec2/imds` to indirect; clean up unused direct deps in `go.mod`/`go.sum`.

<sup>Written for commit 06285d1e38afd38038d8688192fd842bb84d70cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gearnode:
@neilmcgibbon  could you please explain the use case? We don’t want to support configuration outside of the config file. Using AWS LoadDefaultConfig would also introduce environment variable–aware configuration, which we want to avoid.

neilmcgibbon:
> Contributor

@gearnode Sure. The immediate problem was that EKS Pod Identity wasn't working so I can't run this in an EKS environment with Pod Identity. The existing code hardcodes `169.254.170.2` (the ECS metadata endpoint) as the only container credential source, so it silently fails on EKS, which uses 169.254.170.23. That's the bug that prompted this.

On the broader concern, I agree with not wanting arbitrary config sprawl. The difference is that LoadDefaultConfig _isn't_ adding new user-facing configuration; it's delegating credential resolution to the AWS SDK. 

What I'm actually removing is a hand-rolled reimplementation of that chain - one that was already incomplete and had a security hole (unvalidated credential endpoint URLs). I actually fixed that, and extended the manual re-imlemetation of the chain in [this PR](https://github.com/getprobo/probo/pull/1235), but closed that in favour of this one as its in keeping with AWS credential/auth paradigms. The config file API stays _exactly_ the same; AccessKeyID + SecretAccessKey still take priority when set.

The env var point is fair in principle, but the existing code already reads AWS_CONTAINER_CREDENTIALS_RELATIVE_URI directly via os.Getenv. This does not introduce env var sensitivity - its just letting the SDK manage it consistently.

Happy to scope it down if there's a specific concern, but from my end this is a simplification, not a feature addition. Something needed so I can run this in an EKS workload (and I really want to run it, cos its great!)

gearnode:
Agree, the change has been incorporated https://github.com/getprobo/probo/commit/daf139c12d9b44e7c00c74d137944ce64a4061e5

PR #1302, opened 2026-06-05
Add SMTP_HELLO_NAME to fix EHLO rejection by Google Workspace relay

### LINKED ISSUE #1284: SMTP relay fails with Google Workspace because net/smtp defaults EHLO to localhost
## Summary

Self-hosted Probo cannot send mail through Google Workspace SMTP relay (`smtp-relay.gmail.com:587`) because Probo does not let operators configure the SMTP HELO/EHLO identity.

Probo uses Go `net/smtp`, creates the client with `smtp.NewClient(...)`, then calls `StartTLS(...)`. Since Probo never calls `Client.Hello(...)`, Go introduces itself as `localhost` by default. Google Workspace SMTP relay rejects generic HELO/EHLO identifiers such as `localhost`, which causes Probo email sending to fail.

## Environment

- Probo image: `ghcr.io/getprobo/probo:probod-v0.191.0`
- SMTP provider: Google Workspace SMTP relay
- SMTP address: `smtp-relay.gmail.com:587`
- `SMTP_TLS_REQUIRED=true`
- No SMTP username/password; Google relay authenticates by allowed public IP
- Sender address is in the Workspace domain

## Actual Behavior

Queued emails fail with:

```text
cannot send email: TLS negotiation error: EOF
```

This appears as a TLS error, but the failure happens before STARTTLS completes because Go `StartTLS()` performs the initial EHLO automatically.

## Reproduction

A manual SMTP test from the same host succeeds when using a valid EHLO name:

```bash
swaks \
  --server smtp-relay.gmail.com \
  --port 587 \
  --tls \
  --ehlo probo.example.com \
  --from no-reply@example.com \
  --to user@example.com
```

The same relay rejects a generic localhost EHLO:

```bash
swaks \
  --server smtp-relay.gmail.com \
  --port 587 \
  --tls \
  --ehlo localhost \
  --from no-reply@example.com \
  --to user@example.com
```

Observed Google response:

```text
421-4.7.0 Try again later, closing connection. (EHLO)
```

## Root Cause

In `pkg/mailer/mailer.go`, Probo creates an SMTP client and starts TLS:

```go
c, err := smtp.NewClient(conn, host)

if h.smtp.TLSRequired {
    if err := c.StartTLS(&tls.Config{ServerName: host}); err != nil {
        return fmt.Errorf("TLS negotiation error: %w", err)
    }
}
```

Go `net/smtp.NewClient` defaults the local hello name to `localhost`:

```go
c := &Client{Text: text, conn: conn, serverName: host, localName: "localhost"}
```

`StartTLS()` calls the implicit hello path before issuing STARTTLS. Because Probo does not call `c.Hello(...)` first, operators cannot set a valid HELO/EHLO name.

Relevant Go docs:

> The client will introduce itself as "localhost" automatically otherwise.

https://pkg.go.dev/net/smtp#Client.Hello

Google Workspace SMTP relay guidance recommends using a unique HELO/EHLO identifier such as the domain name or server name, and avoiding generic identifiers such as `localhost`.

## Expected Behavior

Self-hosted Probo should allow operators to configure the SMTP HELO/EHLO name so SMTP relays with stricter policies can work.

gearnode:
## Summary

- Adds a new optional `SMTP_HELLO_NAME` config field that operators can set to control the EHLO/HELO identity sent to the SMTP relay.
- When set, `c.Hello(helloName)` is called before `StartTLS`/auth, satisfying strict relay policies (e.g. Google Workspace `smtp-relay.gmail.com` rejects `localhost`).
- When unset, behavior is unchanged (Go's default `localhost` identity is used).

Fixes #1284.

## Changes

- `pkg/probodconfig/mailer_config.go` -- `HelloName` field on `SMTPConfig`
- `pkg/mailer/mailer.go` -- `HelloName` field on `mailer.SMTPConfig`; `c.Hello()` call in `sendMail`
- `pkg/bootstrap/builder.go` -- maps `SMTP_HELLO_NAME` env var
- `pkg/probod/probod.go` -- passes `HelloName` through to the mailer worker
- `pkg/bootstrap/builder_test.go` -- asserts the field defaults to empty
- `contrib/helm/charts/probo/values.yaml` -- `helloName: ""` under `probo.mailer.smtp`
- `contrib/helm/charts/probo/templates/deployment.yaml` -- conditional `SMTP_HELLO_NAME` env var
- `contrib/helm/charts/probo/values-production.yaml.example` -- documents the field with a commented example

## Test plan

- [ ] Bootstrap test passes (`go test ./pkg/bootstrap/`)
- [ ] Mailer package compiles and vets cleanly (`go vet ./pkg/mailer/`)
- [ ] Deploy against `smtp-relay.gmail.com:587` with `SMTP_HELLO_NAME=mail.example.com` and confirm emails are delivered without TLS/EHLO rejection
- [ ] Verify existing deployments without `SMTP_HELLO_NAME` set continue to work unchanged

---
## Summary by cubic
Adds optional `SMTP_HELLO_NAME` to control the EHLO/HELO identity and always send it before TLS/auth to satisfy strict relays like Google Workspace. When unset, the mailer uses `os.Hostname()` and falls back to `localhost`. Fixes #1284.

### Service **`+16`** **`-0`**
- Added `HelloName` to SMTP config and wired through `builder`, `probod`, and `mailer`.
- Mailer calls `c.Hello()` with `SMTP_HELLO_NAME` or `os.Hostname()` (fallback `localhost`) before `StartTLS`.

### Tests **`+1`** **`-0`**
- Asserted default `HelloName` is empty in bootstrap config.

### Other **`+9`** **`-0`**
- Helm: added `helloName` to values, docs, and conditional `SMTP_HELLO_NAME` env in deployment.

<sup>Written for commit 320a86670dbf067abf8002d4e3b8ea9237dc5377. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/mailer/mailer.go
+++ b/pkg/mailer/mailer.go
@@ -49,6 +49,7 @@ type (
 		User        string
 		Password    string
 		TLSRequired bool
+		HelloName   string
 	}
 
 	SendingWorkerOption func(*sendingHandler)
@@ -318,6 +319,12 @@ func (h *sendingHandler) sendMail(ctx context.Context, to []string, msg []byte) 
 	}
 
 	defer func() { _ = c.Quit() }()
+
+	if h.smtp.HelloName != "" {
+		if err := c.Hello(h.smtp.HelloName); err != nil {
+			return fmt.Errorf("SMTP EHLO error: %w", err)
+		}
+	}
 
 	if h.smtp.TLSRequired {
 		if err := c.StartTLS(&tls.Config{ServerName: host}); err != nil {
--- a/pkg/mailer/mailer.go
+++ b/pkg/mailer/mailer.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"net/smtp"
+	"os"
 	"time"
 
 	"github.com/jhillyerd/enmime"
@@ -320,10 +321,15 @@ func (h *sendingHandler) sendMail(ctx context.Context, to []string, msg []byte) 
 
 	defer func() { _ = c.Quit() }()
 
-	if h.smtp.HelloName != "" {
-		if err := c.Hello(h.smtp.HelloName); err != nil {
-			return fmt.Errorf("SMTP EHLO error: %w", err)
+	helloName := h.smtp.HelloName
+	if helloName == "" {
+		if helloName, err = os.Hostname(); err != nil {
+			helloName = "localhost"
 		}
+	}
+
+	if err := c.Hello(helloName); err != nil {
+		return fmt.Errorf("SMTP EHLO error: %w", err)
 	}
 
 	if h.smtp.TLSRequired {
--- a/contrib/helm/charts/probo/templates/deployment.yaml
+++ b/contrib/helm/charts/probo/templates/deployment.yaml
@@ -225,6 +225,10 @@ spec:
             {{- end }}
             - name: SMTP_TLS_REQUIRED
               value: {{ .Values.probo.mailer.smtp.tlsRequired | quote }}
+            {{- if .Values.probo.mailer.smtp.helloName }}
+            - name: SMTP_HELLO_NAME
+              value: {{ .Values.probo.mailer.smtp.helloName | quote }}
+            {{- end }}
             # OpenAI Integration
             {{- if .Values.probo.openai.apiKey }}
             - name: OPENAI_API_KEY
--- a/contrib/helm/charts/probo/values-production.yaml.example
+++ b/contrib/helm/charts/probo/values-production.yaml.example
@@ -157,6 +157,9 @@ probo:
       user: "apikey"
       password: "CHANGE_ME_SMTP_PASSWORD"
       tlsRequired: true
+      # Set to your server's hostname when the relay requires a specific EHLO/HELO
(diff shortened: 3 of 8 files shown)


pkg/mailer/mailer.go
```diff
@@ -319,6 +321,17 @@ func (h *sendingHandler) sendMail(ctx context.Context, to []string, msg []byte)
 
 	defer func() { _ = c.Quit() }()
 
+	helloName := h.smtp.HelloName
+	if helloName == "" {
+		if helloName, err = os.Hostname(); err != nil {
```
lukkor, line 326: I don't think that it is preserving previous behavior if you are using `os.Hostname()`; if `helloName == ""`, the call to the function hello should not be executed at all.

PR #1331, opened 2026-06-10
Make keyset cursor pagination NULL-aware

SachaProbo:
Cursor.SQLFragment emitted a seek predicate with no NULL handling, so once a page boundary landed on a NULL order value the WHERE matched no rows and pagination terminated early, silently dropping the remaining rows.

Rewrite SQLFragment to emit explicit NULLS placement and a null-aware, type-agnostic seek predicate, mirroring the direction and NULLS placement on the Head/Tail inversion so backward paging stays an exact inverse. For NOT NULL columns it reduces to the previous predicate.

Add white-box pagination tests that evaluate the generated SQL with three-valued logic and walk the whole set forward and backward across the NULL boundary.

---
## Summary by cubic
Makes keyset cursor pagination NULL-aware so pages don’t truncate when the sort column has NULLs. Adds explicit NULLS placement and a null-aware seek; backward paging mirrors forward exactly.

### Tests **`+785`** **`-0`**
- Add Postgres-backed e2e walk over adversarial datasets (NULL blocks, ties) to confirm forward/backward paging reconstruct full order.
- Lock exact SQL and named args across direction/position/key cases; `@cursor_field_value` appears only for non-NULL seeks.
- Add unit tests for `CursorKey` round-trip and invalid formats, `OrderDirection` text marshaling, and `Page` trimming/HasNext/HasPrev for Head/Tail.

### Service **`+67`** **`-21`**
- Rewrite `Cursor.SQLFragment` via a template that sets ASC/DESC, NULLS FIRST/LAST, and id tiebreak; Tail flips both for symmetry.
- Use a null-aware, type-agnostic seek; bind `@cursor_field_value` only when non-NULL.

### Other **`+2`** **`-2`**
- Run the new e2e suite in CI and Makefile by including `./e2e/page/...`.

<sup>Written for commit 00d4d5f62f5561900dde09fb57a9cb967519282f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1331?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/page/cursor.go
+++ b/pkg/page/cursor.go
@@ -15,6 +15,9 @@
 package page
 
 import (
+	"strings"
+	"text/template"
+
 	"github.com/jackc/pgx/v5"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/page/cursor.go
```diff
@@ -15,6 +15,9 @@
 package page
 
 import (
+	"strings"
```
gearnode, line 18: add test with real pg

PR #1348, opened 2026-06-12
Redirect to the org SAML login on SSO-required assume

SachaProbo:
When assuming an organization with a REQUIRED SAML policy, a non-SAML
session must re-authenticate through that org's IdP. The console derived
the SAML URL from the viewer's email domain (viewer.ssoLoginURL), which
errors for domains with zero or multiple SAML configs, so the assume
page threw and bounced the user to the email/password screen.

Now ErrSAMLAuthenticationRequired carries the org's SAML configuration
ID, the resolver returns the org-specific ssoLoginURL in the
SAMLAuthenticationRequired result, and the console redirects straight to
it. SAML enforcement stays strict: only the org's own SAML session
satisfies a REQUIRED policy.

---
## Summary by cubic
Redirect users to the organization’s SAML login when assuming an org with SAML set to REQUIRED. Uses the org-specific SSO URL to avoid domain-based failures and ensure only the org’s SAML session satisfies the policy.

### Service **`+5`** **`-4`**
Add SAML configuration ID to `ErrSAMLAuthenticationRequired` and update `NewSAMLAuthenticationRequiredError`. In `AssumeOrganizationSession`, when SAML is REQUIRED and the root session isn’t SAML, return this error with the org’s config ID to force reauth via the org IdP.

### GraphQL API **`+3`** **`-1`**
Expose `ssoLoginURL` on `SAMLAuthenticationRequired` and resolve it using the carried config ID.

### App: console **`+8`** **`-14`**
Assume page redirects using `result.ssoLoginURL` and attaches a continue param. Remove `viewer.ssoLoginURL` and query-string reliance to prevent redirect errors.

<sup>Written for commit a0a71f46493c777242790266e2da6d0f51a1c201. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/console/src/pages/iam/organizations/AssumePage.tsx
+++ b/apps/console/src/pages/iam/organizations/AssumePage.tsx
@@ -16,7 +16,7 @@ import { useTranslate } from "@probo/i18n";
 import { UnAuthenticatedError } from "@probo/relay";
 import { useEffect } from "react";
 import { type PreloadedQuery, useMutation, usePreloadedQuery } from "react-relay";
-import { useNavigate, useSearchParams } from "react-router";
+import { useNavigate } from "react-router";
 import { graphql } from "relay-runtime";
 
 import type { AssumePageMutation } from "#/__generated__/iam/AssumePageMutation.graphql";
@@ -38,6 +38,7 @@ const assumeMutation = graphql`
         }
         ... on SAMLAuthenticationRequired {
           reason
+          ssoLoginURL
         }
       }
     }
@@ -48,9 +49,6 @@ export const assumePageQuery = graphql`
   query AssumePageQuery {
     viewer @required(action: THROW) {
       __typename
-      ... on Identity {
-        ssoLoginURL
-      }
     }
   }
 `;
@@ -60,12 +58,11 @@ export function AssumePage(props: { queryRef: PreloadedQuery<AssumePageQuery> })
 
   const organizationId = useOrganizationId();
   const navigate = useNavigate();
-  const [searchParams] = useSearchParams();
   const { __ } = useTranslate();
 
   const safeContinueUrl = useSafeContinueUrl(`/organizations/${organizationId}`);
 
-  const { viewer } = usePreloadedQuery<AssumePageQuery>(assumePageQuery, queryRef);
+  usePreloadedQuery<AssumePageQuery>(assumePageQuery, queryRef);
   const [assumeOrganizationSession] = useMutation<AssumePageMutation>(assumeMutation);
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


pkg/iam/session_service.go
```diff
@@ -580,7 +580,16 @@ func (s SessionService) AssumeOrganizationSession(
...
+					case coredata.AuthMethodSAML,
+						coredata.AuthMethodOIDC:
+						// SAML and OIDC (Google / Microsoft) are both federated
+						// SSO methods, so either satisfies an organization that
+						// requires SSO. A Google / Microsoft sign-in is valid
+						// across every organization the user belongs to.
+					default:
+						// PASSWORD and MAGIC_LINK are not federated SSO, so the
+						// user must re-authenticate through SSO.
 						return NewSAMLAuthenticationRequiredError("policy_requirement")
```
codenem: should we rename this error then?

pkg/iam/session_service.go
```diff
@@ -580,7 +580,16 @@ func (s SessionService) AssumeOrganizationSession(
 			if err == nil {
 				switch samlConfig.EnforcementPolicy {
 				case coredata.SAMLEnforcementPolicyRequired:
```
codenem, line 582: And this one?

codenem (approved):

gearnode (changes requested):
This is not correct, Google or Microsoft SSO are not SAML auth.

apps/console/src/pages/iam/organizations/AssumePage.tsx
```diff
@@ -60,12 +58,11 @@ export function AssumePage(props: { queryRef: PreloadedQuery<AssumePageQuery> })
 
   const organizationId = useOrganizationId();
   const navigate = useNavigate();
-  const [searchParams] = useSearchParams();
   const { __ } = useTranslate();
 
   const safeContinueUrl = useSafeContinueUrl(`/organizations/${organizationId}`);
 
-  const { viewer } = usePreloadedQuery<AssumePageQuery>(assumePageQuery, queryRef);
+  usePreloadedQuery<AssumePageQuery>(assumePageQuery, queryRef);
```
codenem, line 65: seems we don't use any data here

apps/console/src/pages/iam/organizations/AssumePage.tsx
```diff
@@ -48,9 +49,6 @@ export const assumePageQuery = graphql`
   query AssumePageQuery {
     viewer @required(action: THROW) {
       __typename
-      ... on Identity {
-        ssoLoginURL
```
codenem, line 52: maybe we don't need this in the gql schema anymore

codenem (approved):
apart some potential cleaning LGTM

PR #1471, opened 2026-07-07
Resolve document publish approvers from defaults

SachaProbo:
The MCP publishDocument tool required callers to pass approver_ids and to distinguish an omitted list (rejected) from an empty one (direct publish), a null-vs-empty subtlety that is awkward for an LLM to get right.

Drop approver_ids from the MCP tool and resolve a major publish's approvers from the document's default approvers instead: an approval is requested when the document has default approvers, otherwise the version is published directly. Default approvers are configured with addDocument or updateDocument.

Replace a document's default approvers on every major publish that supplies an explicit list, even when the list is empty, so a direct publish through the GraphQL API clears stale approvers instead of leaving them behind (previously the empty case skipped the update).

Expose the default-approver behaviour as a separate entry point, PublishVersionWithDefaultApprovers, that loads the defaults and delegates to PublishVersion. PublishVersion keeps its explicit-approver contract for the GraphQL API.

Require only the publish permission to publish a version, whether or not it opens an approval quorum, and drop the now-unused request-approval action.

Fold the publish steps into the publishMinor and publishMajor primitives shared by both the single and bulk publish paths, and drop the redundant InTx suffix from RequestApproval and emitDocumentEvent, which already take a transaction argument.

---
## Summary by cubic
MCP major publishes now resolve approvers from a document’s defaults: if defaults exist we open an approval quorum; if not, we publish directly. GraphQL still supports explicit approvers, and major publishes update the document’s default approvers (empty clears), with only the publish permission required.

### Service **`+269`** **`-169`**
- Added `PublishVersionWithDefaultApprovers` to load defaults and delegate to `PublishVersion`.
- Major publishes always update default approvers from the provided list (including empty).
- Extracted `publishMajor`/`publishMinor` to handle single and bulk paths and emit version-published events.
- Dropped request-approval action and its OAuth scope; publishing always authorizes with `ActionDocumentVersionPublish`.
- Renamed `RequestApprovalInTx`→`RequestApproval` and `emitDocumentEventInTx`→`emitDocumentEvent`; updated call sites (generated flows, bulk publish, others).

### MCP **`+7`** **`-18`**
- `publishDocument` tool no longer accepts `approver_ids`; it resolves from document defaults via `PublishVersionWithDefaultApprovers`.
- Updated `specification.yaml` to reflect simpler input and default-approver behavior.

### GraphQL API **`+1`** **`-6`**
- `publishDocument` now always authorizes with the publish action only.

### Tests **`+107`** **`-0`**
- Added MCP e2e tests for approval quorum when defaults exist and direct publish when not.
- Added console e2e test verifying that an explicit empty approver list on a major publish clears defaults and publishes.

<sup>Written for commit 936162d5c770e08e2661ea0192449e487ea445a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1471?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/e2e/console/document_version_test.go
+++ b/e2e/console/document_version_test.go
@@ -1688,6 +1688,45 @@ func TestDocument_DefaultApprovers(t *testing.T) {
 
 		assert.Empty(t, result.UpdateDocument.Document.DefaultApprovers)
 	})
+
+	t.Run("major publish with explicit approvers updates default approvers, even when empty", func(t *testing.T) {
+		t.Parallel()
+
+		docID := createTestDocumentWithApprovers(t, owner, []string{approverID})
+
+		var result struct {
+			PublishDocument struct {
+				Document struct {
+					DefaultApprovers []struct {
+						ID string `json:"id"`
+					} `json:"defaultApprovers"`
+				} `json:"document"`
+				DocumentVersion struct {
+					Status string `json:"status"`
+				} `json:"documentVersion"`
+			} `json:"publishDocument"`
+		}
+
+		err := owner.Execute(`
+			mutation($input: PublishDocumentInput!) {
+				publishDocument(input: $input) {
+					document { defaultApprovers { id } }
+					documentVersion { status }
+				}
+			}
+		`, map[string]any{
+			"input": map[string]any{
+				"minor":       false,
+				"documentId":  docID,
+				"approverIds": []string{},
+				"changelog":   "Direct publish clearing approvers",
+			},
+		}, &result)
+		require.NoError(t, err)
+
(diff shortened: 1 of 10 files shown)


lukkor (approved):
LGTM

PR #1479, opened 2026-07-10
fix(cli): third-party list/update schema drift + kebab-case command name

uvollmer:
## Summary

Two `prb thirdParty` subcommands still use snake_case field names that the console schema no longer exposes, so they fail on every invocation with `GRAPHQL_VALIDATION_FAILED`:

- **list** queried `Organization.third_parties` — the schema exposes `thirdParties` (`pkg/server/api/console/v1/graphql/organization.graphql`)
- **update** selected `UpdateThirdPartyPayload.third_party` — the schema exposes `thirdParty`

Both the GraphQL documents and the response-decoding struct tags are aligned with the schema. Also updates stale `prb third_party ...` example texts to the actual command name (`prb thirdParty ...`).

## Testing

- Reproduced both 422s against a live instance; verified the fixed build lists and updates third parties correctly
- `go build ./pkg/cmd/...` and `go vet` pass

### DIFF
--- a/pkg/cmd/thirdpartymgmt/create/create.go
+++ b/pkg/cmd/thirdpartymgmt/create/create.go
@@ -63,12 +63,12 @@ func NewCmdCreate(f *cmdutil.Factory) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "create",
-		Short: "Create a new thirdParty",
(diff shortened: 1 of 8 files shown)


gearnode (changes requested):
use kebab case please.

PR #1583, opened 2026-07-30
Reuse existing membership profiles when creating users

SachaProbo:
Allows promoting compliance-portal or orphan profiles into console members instead of failing on the unique identity/org constraint. Adds e2e coverage for the reuse path.

---
## Summary by cubic
Reuses existing compliance-portal or orphan membership profiles when creating users to avoid identity/org conflicts and promote them into console members. Adds e2e coverage and returns clear conflicts for SCIM-managed or duplicate users.

### Tests **`+193`** **`-0`**
- Seeds an orphan profile, runs createUser, and verifies the profile ID is reused.
- Confirms the orphan is hidden until membership exists, preserves state, sets role, and a second create returns CONFLICT.

### Service **`+93`** **`-23`**
- Loads a profile by identity+organization; if found without a membership, reuses it and updates fields (name, kind, emails, dates) while keeping state. Returns a specific error for SCIM-managed profiles.
- Maps duplicate profile/membership inserts to “user already exists”.
- When contractEndDate is in the past, deletes requested document signatures for the profile.

### GraphQL API **`+4`** **`-0`**
- Maps SCIM-managed user errors to a Conflict with a clear “user is managed by SCIM and cannot be created manually” message.

### MCP **`+4`** **`-0`**
- Returns a clear “user managed by SCIM” message on create instead of a generic error.

<sup>Written for commit e4a35a503c26da2eeb2dd0cefb4225660e63e331. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1583?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/iam/organization_service.go
+++ b/pkg/iam/organization_service.go
@@ -1036,36 +1036,102 @@ func (s *OrganizationService) CreateUser(ctx context.Context, scope coredata.Sco
 				}
 			}
 
-			profile = &coredata.MembershipProfile{
-				ID:                       gid.New(req.OrganizationID.TenantID(), coredata.MembershipProfileEntityType),
-				IdentityID:               identity.ID,
-				OrganizationID:           req.OrganizationID,
-				EmailAddress:             req.EmailAddress,
-				Source:                   coredata.ProfileSourceManual,
-				FullName:                 req.FullName,
-				Kind:                     req.Kind,
-				AdditionalEmailAddresses: req.AdditionalEmailAddresses,
-				Position:                 req.Position,
-				// User is pending until they accept an invitation.
-				State:     coredata.ProfileStatePending,
-				CreatedAt: now,
-				UpdatedAt: now,
+			existingProfile := &coredata.MembershipProfile{}
+
+			err := existingProfile.LoadByIdentityIDAndOrganizationID(
+				ctx,
+				conn,
+				scope,
+				identity.ID,
+				req.OrganizationID,
+			)
+			if err != nil && !errors.Is(err, coredata.ErrResourceNotFound) {
+				return fmt.Errorf("cannot load profile: %w", err)
 			}
 
-			if req.ContractStartDate != nil {
-				profile.ContractStartDate = *req.ContractStartDate
-			}
+			if err == nil {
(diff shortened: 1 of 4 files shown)


pkg/iam/organization_service.go
```diff
@@ -1036,36 +1036,102 @@ func (s *OrganizationService) CreateUser(ctx context.Context, scope coredata.Sco
...
+
+				if req.ContractEndDate != nil {
+					profile.ContractEndDate = *req.ContractEndDate
+				}
+
+				if err := profile.Update(ctx, conn, scope); err != nil {
+					return fmt.Errorf("cannot update profile: %w", err)
+				}
+
+				if profile.ContractEndDate != nil && profile.ContractEndDate.Before(now) {
```
codenem, line 1097: Why do we do this here? There could be other operations to do when a contract has ended. We have a linear card to deactivate users in a worker when contract date is done

pkg/iam/organization_service.go
codenem, line 1007: Maybe we can be super clear here and rename `UpsertUserAndCreateMembership`

pkg/iam/organization_service.go
```diff
@@ -1036,36 +1036,102 @@ func (s *OrganizationService) CreateUser(ctx context.Context, scope coredata.Sco
...
+				if err == nil {
 					return NewUserAlreadyExistsError(identity.ID, req.OrganizationID)
 				}
 
-				return fmt.Errorf("cannot insert profile: %w", err)
+				if !errors.Is(err, coredata.ErrResourceNotFound) {
+					return fmt.Errorf("cannot load membership: %w", err)
+				}
+
+				// Reuse orphan / compliance-portal profiles so the same person
```
codenem, line 1074: How can a profile be orphan?

PR #1597, opened 2026-07-30
Default MCP third-party list to level 1

SachaProbo:
Agents usually need direct vendors, not the full hierarchy. Mirror listDocuments: omit level to get level 1, pass level to override. Document the default in the tool and field descriptions.

---
## Summary by cubic
Default `listThirdParties` to return only level 1 (direct third parties) when `level` is omitted. This mirrors `listDocuments` and avoids fetching the full hierarchy by default.

### MCP **`+8`** **`-3`**
- In `schema.resolvers.go`, set `level` to 1 when `input.Level` is nil before calling `coredata.NewThirdPartyFilter`.
- In `specification.yaml`, define `level` as hierarchy depth with a default of 1 and clarify the `listThirdParties` tool description.

<sup>Written for commit 41da4bbad12b283780161cc4ef71aaebf91e7e7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/server/api/mcp/v1/specification.yaml
+++ b/pkg/server/api/mcp/v1/specification.yaml
@@ -663,7 +663,7 @@ components:
           description: Organization ID
         level:
           type: integer
-          description: Filter by level
+          description: Filter by hierarchy depth (1 = direct third party, 2+ = third party of a third party / indirect). Defaults to 1 when omitted.
         order_by:
           $ref: "#/components/schemas/ThirdPartyOrderBy"
           description: ThirdParty order by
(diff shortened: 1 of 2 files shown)


pkg/server/api/mcp/v1/specification.yaml
```diff
@@ -663,7 +663,7 @@ components:
           description: Organization ID
         level:
           type: integer
-          description: Filter by level
+          description: Filter by level. Defaults to 1 when omitted.
```
gearnode: maybe we should explain better what levels are.

PR #1604, opened 2026-07-31
Support compliance portal custom domains behind a TLS-terminating proxy

GabrielHogan:
This was mostly a test to see if claude could accomplish this fix for me. I honestly didn't mean to make a PR, but here we are :) 

Let's see what the AI review thinks. If it's helpful at all then great, if not I would just love a way to serve the compliance portal on a custom domain without ACME or CNAME verification due to it being behind a cloudflare tunnel.

Thanks!

----- ALL OF THE FOLLOWING IS AI GENERATED WITH VERY BASIC HUMAN REVIEW -----

## Problem

A self-hosted Probo behind a reverse proxy, tunnel or CDN that terminates public TLS — Cloudflare Tunnel, CloudFront, an ingress controller with its own certificates — cannot use compliance portal custom domains at all. Two independent reasons:

**1. The CNAME pre-check can never pass.** `beginChallengeHandler.begin()` (`pkg/certmanager/provision_worker.go`) runs `dnsClient.CheckCNAME(hostname, cnameTarget)` before placing an ACME order. A hostname routed through a proxying CDN resolves to the CDN edge in public DNS — the origin CNAME is not visible there, and for Cloudflare Tunnel the record *has* to be proxied for routing to work at all. So the check fails forever, even when HTTP routing to the deployment is correct and the HTTP-01 challenge would succeed:

```
{"level":"WARN","msg":"dns configuration check failed","domain":"trust.example.com",
 "error":"no cname records found for domain \"trust.example.com\""}
```

`skipsDNSChecks` is the only escape hatch and only covers `managed=true` rows, which no user-facing path creates (`Service.AddCustomDomain` hardcodes `false`; the only `managed=true` domain is the auto-created probopage subdomain).

**2. Even with a certificate, the portal insists on terminating TLS itself.** `runCompliancePortalServer` (`pkg/probod/probod.go`) unconditionally 301-redirects `:80` to HTTPS and selects a Probo-issued certificate on `:443`. When the tunnel already terminates public TLS, that certificate is redundant, and there is no way to opt out. `ProxyProtocol.TrustedProxies` only wires literal PROXY-protocol framing, which Cloudflare Tunnel does not speak.

## What this PR does

**Option A (primary) — `PROBOD_TRUST_CENTER_TLS_TERMINATED_BY_PROXY`** (`probo.compliancePortal.tlsTerminatedByProxy`, default `false`).

| | before | with the flag |
|---|---|---|
| `:80` | ACME challenges + 301 to HTTPS | serves the portal router directly |
| `:443` | portal with a Probo-issued cert | not started |
| request → page | resolved from the TLS SNI | resolved from the `Host` header (port stripped) |
| advertised URLs | `https://` | `https://` (unchanged — the proxy speaks TLS to the world) |
| certificate manager | provisions and renews | workers not started, no ACME order placed |
| custom domain serves | once its certificate is `ACTIVE` | as soon as it is added |

**Option C (complement) — `PROBOD_CUSTOM_DOMAINS_SKIP_CNAME_CHECK`** (`probo.customDomains.skipCnameCheck`, default `false`) for operators who *do* want a Probo-managed certificate but sit behind a proxying CDN. It downgrades the CNAME pre-check to a warning and lets the ACME HTTP-01 challenge — which is what actually proves control of the hostname — decide. The CAA check still blocks, since a CAA policy that forbids the CA guarantees the order fails.

**Option B (BYO certificate)** is not implemented. With Option A there is no Probo-terminated TLS to supply a certificate for, so it would only serve operators who want Probo to terminate TLS with their own material — a reasonable follow-up, but not needed to unblock this class of deployment.

### Why the CNAME check was not relaxed unconditionally

Attempting HTTP-01 for every domain regardless would make misconfigured domains consume ACME rate limits, and after `maxProvisioningRetries` they would go terminal `FAILED` instead of staying `PENDING` and auto-recovering when DNS is fixed. That is a real regression for a shared install, so the relaxation is opt-in and server-level.

## Security note (multi-tenant)

`custom_domains.domain` is globally unique across tenants, and domain-ownership verification is what stops one tenant claiming a hostname it does not control.

- **`TLS_TERMINATED_BY_PROXY` does remove that verification.** With no certificate there is no ACME challenge; a registered domain is treated as verified, both for serving and for the safe-redirect allowlist (`IsVerifiedRedirectHost`). The operator vouches for the hosts their proxy routes to Probo. **Single/trusted-tenant, self-hosted deployments only.**
- **`SKIP_CNAME_CHECK` does not.** HTTP-01 still proves control; what is lost is the cheap upfront filter, so a tenant could make the server spend ACME attempts on hostnames it does not control.

Both are **server-level env vars, not GraphQL-mutable per-organization settings**, so no tenant can enable them on a shared install. The tradeoffs are documented in `contrib/helm/charts/probo/values.yaml`, `contrib/helm/README.md` and both changelogs.

## Testing

- `pkg/certmanager/provision_worker_test.go` — new `checkDNSConfiguration` cases behind a small `dnsChecker` interface (stubbed): CNAME mismatch blocks by default, is a hint when skipped, CAA still blocks when the CNAME check is skipped, both checks run in the default path.
- `pkg/server/api/complianceportal/sni_middleware_test.go` (new) — `portalHostFromRequest`: TLS SNI wins, plain HTTP carries no portal host by default, `Host` (port stripped) is used behind a TLS-terminating proxy.
- `pkg/probod/compliance_portal_test.go` (new) — the `:80` listener serves the portal router with the flag on and the ACME/redirect handler without it.
- `pkg/bootstrap/builder_test.go` — env-var mapping and defaults for both flags.
- `go build ./...`, `go vet ./...`, `gofmt -l apps cmd packages pkg e2e`, and `go test` for the touched packages all pass. Defaults are unchanged, so existing TLS-terminating/ACME behaviour keeps its current coverage.

## Known limitation / follow-up

The console's domain card still renders the certificate status badge, so a working domain reads "Pending" in `TLS_TERMINATED_BY_PROXY` mode. Surfacing the mode in the console would need a new server-config field on the GraphQL API; it is deliberately left out of this change, which adds no UI.

## Config propagation

Per `contrib/claude/config.md`: `pkg/probodconfig/`, `pkg/bootstrap/builder.go` (+ tests), `contrib/helm/charts/probo/values.yaml`, `values-production.yaml.example`, `templates/deployment.yaml`, `contrib/helm/README.md`, `cmd/probod/CHANGELOG.md`, `contrib/helm/charts/probo/CHANGELOG.md`. No secret material, so `secret.yaml` is untouched; both defaults match the bootstrap defaults, so `provision.sh` and the e2e env map need no entry.

Both commits are DCO signed off.

---
## Summary by cubic
Adds support for hosting the compliance portal behind a TLS-terminating proxy/CDN. Adds flags to serve the portal on HTTP without Probo-managed TLS and to optionally skip the CNAME pre-check so ACME can work behind CDNs; URLs stay https.

### Service **`+271`** **`-124`**
- New env/config: `PROBOD_TRUST_CENTER_TLS_TERMINATED_BY_PROXY`, `PROBOD_CUSTOM_DOMAINS_SKIP_CNAME_CHECK`.
- Proxy mode: don’t start HTTPS or certificate workers; serve the portal on `:80`; domains serve immediately; verification uses registration only.
- Cert manager: `dnsChecker` and `checkDNSConfiguration`; CNAME pre-check becomes a warning when opted-in; CAA still blocks.
- Wiring in `pkg/bootstrap` and `pkg/probodconfig`; listener/shutdown helpers extracted.

### GraphQL API **`+148`** **`-84`**
- SNI middleware resolves host from TLS SNI or `Host` when TLS is terminated upstream; strips port; base URLs stay `https://`.
- `pkg/server/api/complianceportal/v1/mux.go` passes the proxy mode flag.

### Tests **`+280`** **`-0`**
- DNS pre-check skip and CAA blocking.
- Host resolution and canonical redirects.
- HTTP handler selection (portal vs ACME/redirect) and env mapping defaults.

### Other **`+93`** **`-0`**
- Helm values, deployment envs, and README for both flags.
- Changelogs updated.

<sup>Written for commit 323de69a4f2886482c122ef234a60073e0e3371e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1604?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lukkor:
Thanks @GabrielHogan for your contribution. Similar work has just been merged to `main` and will be released shortly.

PR #1613, opened 2026-08-01
transfermanager: honour the S3 client's RequestChecksumCalculation

anthonymorrisjohnson:
## Problem

`PutFile` builds the S3 transfer manager with no options:

```go
uploader := transfermanager.New(s.s3Client)
```

The transfer manager keeps its **own** `RequestChecksumCalculation`, and its documentation is explicit that it *"takes precedence over the `RequestChecksumCalculation` configured on the underlying S3 client"*. It defaults to `WhenSupported`.

So the client's setting is silently discarded. A deployment that configures `when_required` — via `AWS_REQUEST_CHECKSUM_CALCULATION`, shared config, or `aws.Config` directly — still gets a CRC32 checksum on every transfer-manager upload.

## Evidence

Captured the actual request headers the transfer manager emits against an `httptest` server, with `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` set:

**Before (current `main`)**
```
Content-Length:                 13
X-Amz-Checksum-Crc32:           Ww2ACA==
X-Amz-Content-Sha256:           1b5bfa2f…
X-Amz-Sdk-Checksum-Algorithm:   CRC32
```

**After**
```
Content-Length:                 13
X-Amz-Content-Sha256:           1b5bfa2f…
```

The configuration is requested but not applied; after the change it is.

## Why it matters

Some S3-compatible endpoints do not implement AWS's checksum scheme and reject writes that carry it. `when_required` is the documented escape hatch for exactly that, and today it has no effect on this code path — leaving no way to turn the behaviour off short of patching.

## Fix

Propagate the client's setting to the transfer manager, so configuring it takes effect:

```go
uploader := transfermanager.New(s.s3Client, func(o *transfermanager.Options) {
    o.RequestChecksumCalculation = s.s3Client.Options().RequestChecksumCalculation
})
```

## Compatibility

No behaviour change for anyone who has not set it — the client and the transfer manager both default to `WhenSupported`, so the propagated value is identical. Only deployments that explicitly configured `when_required` see a difference, and that difference is what they asked for.

## Scope — deliberately narrow

This fixes `filemanager.PutFile` only. Note that several other upload paths (`pkg/probo/file_service.go`, `document_service.go`, `framework_service.go`, the `third_party_*_agreement` services, and `pkg/complianceportal/management/*`) call `s3.PutObject` directly rather than going through `filemanager`, so they are unaffected by this change — they already honour the client's setting, since they use the client directly. I have not touched them; consolidating those onto `filemanager` looks worthwhile but is a separate concern and a much larger diff.

## Tests

Adds `TestPutFile_HonoursRequestChecksumCalculation`, a table test asserting the setting is honoured in **both** directions — `when_required` produces no checksum, `when_supported` still produces one. It uses the existing `httptest` harness, so no external dependency, and inspects the request for any of the forms the SDK can use (declared algorithm, precomputed header, or chunked body with trailing checksum).

Verified as a real regression guard, not merely a passing test:

- with the fix: both cases pass
- with the fix reverted: `when_required` fails (`expected: false, actual: true`), `when_supported` still passes

`gofmt` and `go vet` are clean, and the full `pkg/filemanager` suite passes.

### DIFF
--- a/pkg/filemanager/s3.go
+++ b/pkg/filemanager/s3.go
@@ -234,7 +234,19 @@ func (s *Service) PutFile(
 	// Transfer manager accepts unseekable readers (e.g. io.Pipe) by buffering
 	// parts in memory, which works against plain-HTTP S3-compatible endpoints
 	// where PutObject checksums require a seekable body.
-	uploader := transfermanager.New(s.s3Client)
+	//
+	// The transfer manager keeps its OWN RequestChecksumCalculation, which takes
+	// precedence over the one configured on the S3 client and defaults to
+	// WhenSupported. Left alone it therefore discards the client's setting —
+	// including AWS_REQUEST_CHECKSUM_CALCULATION and the shared-config value —
+	// and adds a CRC32 checksum to every upload regardless. S3-compatible
+	// endpoints that do not implement AWS's checksum scheme reject those writes,
+	// and the failure surfaces as an opaque `SignatureDoesNotMatch: Access
+	// denied`, which reads as a credential problem rather than a checksum one.
+	// Propagate the client's setting so configuring it actually takes effect.
+	uploader := transfermanager.New(s.s3Client, func(o *transfermanager.Options) {
+		o.RequestChecksumCalculation = s.s3Client.Options().RequestChecksumCalculation
+	})
 
 	input := &transfermanager.UploadObjectInput{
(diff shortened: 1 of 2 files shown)


gearnode:
Thanks!

PR #1686, opened 2026-08-11
fix(esign): print signed_at on certificates at seal precision

### LINKED ISSUE #1684: Certificate of Completion prints Signed At at second precision, so the documented seal verification always fails
## Summary

The Certificate of Completion renders `Signed At` at second precision, but the integrity seal is computed over the same timestamp at microsecond precision. The certificate's own "Seal Verification (v1)" section instructs the reader to recompute the seal from the fields shown on the certificate — but the timestamp shown is not the timestamp that was hashed, so following those instructions always produces a mismatch.

Observed on v0.216.1.

## Cause

`computeSealV1` hashes the timestamp at microsecond precision:

```go
// pkg/coredata/electronic_signature.go
es.SignedAt.UTC().Truncate(time.Microsecond).Format(time.RFC3339Nano),
```

`certgen` renders it at second precision:

```go
// pkg/esign/certgen.go:100
data.SignedAt = signature.SignedAt.UTC().Format(time.RFC3339)
```

So a certificate showing `2026-01-02T03:04:05Z` may have been sealed over `2026-01-02T03:04:05.961506Z`. The sub-second component appears nowhere on the certificate, and `ElectronicSignatureID` is `json:"-"`, so it isn't reachable through the API either. It exists only in `electronic_signatures.signed_at` in the database.

## Impact

Independent verifiability is the stated purpose of the Seal Verification section, and it's the property that makes the seal meaningful to a third party — an auditor, counterparty, or anyone checking a signature without database access. As shipped, that verification cannot succeed.

The failure mode is also the worst-looking one: a verifier who follows the printed instructions gets a mismatched digest, which reads as tampering rather than as a formatting bug.

The seal itself is correct. Recomputing across all 10^6 microsecond values recovers the exact timestamp and produces a matching digest, which confirms both the diagnosis and that the underlying seal is sound.

## Suggested fix

Render `SignedAt` on the certificate using the same expression the seal uses, so the printed value is byte-identical to the hashed one:

```go
data.SignedAt = signature.SignedAt.UTC().Truncate(time.Microsecond).Format(time.RFC3339Nano)
```

Existing certificates would remain unverifiable by the printed procedure, since the precision was never persisted into them; only newly generated ones would be fixed.

AnotherDork:
Fixes #1684.

## Problem

`certgen` rendered the signing timestamp with `time.RFC3339` (second precision) while `computeSealV1` hashes it truncated to microseconds. The certificate's own "Seal Verification (v1)" section instructs the reader to recompute the seal from the fields shown on the certificate, so whenever the fraction was non-zero that procedure could not succeed — the timestamp printed was not the timestamp hashed.

The failure mode is worse than an unhelpful one: a verifier following the printed steps gets a mismatched digest, which reads as evidence of tampering rather than as a formatting difference. Independent verifiability is the property the seal section exists to provide.

## Change

Adds `ElectronicSignature.SealSignedAt()` as the single rendering of that timestamp, used by both `computeSealV1` and `certgen`, so the displayed value and the hashed value cannot drift apart again. I went with a shared accessor rather than duplicating the format expression in `certgen`, since the duplication is what allowed the two to diverge in the first place — happy to reduce it to a one-line change in `certgen` if you'd prefer the smaller diff.

No behaviour change to the seal itself: the hashed bytes are identical before and after.

## Tests

`pkg/coredata/electronic_signature_seal_test.go` asserts that what `SealSignedAt()` returns is exactly what the seal hashes, across the formatting cases that matter — microsecond fractions, trailing-zero stripping (`.1Z`, not `.100000Z`), a whole second carrying no fraction at all, and nanosecond input truncated to microseconds — plus the nil case.

Verified against a real certificate from a self-hosted instance: the seal recomputes and matches once the full-precision timestamp is used.

## Note for existing certificates

Certificates generated before this change stay unverifiable by the printed procedure, because the sub-second component was never written into them. Recovering it means recomputing across the 10^6 candidate microsecond values, which is deterministic and takes about a second. Worth a mention in release notes if any users are relying on that section.

---
## Summary by cubic
Unifies certificate and seal timestamp rendering by introducing a single `SealSignedAt()` formatter and using it in both the seal and certificate. This fixes seal verification when sub-second fractions exist and avoids false tampering warnings.

### Tests **`+127`** **`-0`**
Adds tests ensuring `SealSignedAt()` is exactly what the seal hashes, covering microseconds, trailing-zero stripping, whole seconds, nanosecond truncation, and nil.

### Coredata **`+12`** **`-1`**
Adds `ElectronicSignature.SealSignedAt()` that returns `SignedAt` in UTC, truncated to microseconds, using `time.RFC3339Nano`. Uses it inside `computeSealV1` so the hashed and displayed timestamps stay in sync.

### Service **`+1`** **`-1`**
Updates `pkg/esign/certgen.go` to print `data.SignedAt = signature.SealSignedAt()` so certificates show the same timestamp that the seal hashes.

<sup>Written for commit ef7b9190a5d83ab5cf48ec6f64348adfdce94bfc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1686?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/coredata/electronic_signature.go
+++ b/pkg/coredata/electronic_signature.go
@@ -373,6 +373,17 @@ func (es *ElectronicSignature) ComputeSeal(version int) (string, error) {
 	}
 }
 
+// SealSignedAt renders SignedAt exactly as the seal hashes it. Anything
+// displaying the timestamp to a verifier must use this, or the seal cannot be
+// recomputed from what they were shown.
+func (es *ElectronicSignature) SealSignedAt() string {
+	if es.SignedAt == nil {
+		return ""
+	}
+
+	return es.SignedAt.UTC().Truncate(time.Microsecond).Format(time.RFC3339Nano)
+}
+
 func (es *ElectronicSignature) computeSealV1() (string, error) {
 	if es.SignedAt == nil {
 		return "", fmt.Errorf("signed_at must not be nil")
@@ -389,7 +400,7 @@ func (es *ElectronicSignature) computeSealV1() (string, error) {
 		ref.UnrefOrZero(es.SignerIPAddress),
 		ref.UnrefOrZero(es.SignerUserAgent),
 		es.ConsentText,
-		es.SignedAt.UTC().Truncate(time.Microsecond).Format(time.RFC3339Nano),
+		es.SealSignedAt(),
 	}
 
 	for i, f := range fields {
--- a/pkg/coredata/electronic_signature_seal_test.go
+++ b/pkg/coredata/electronic_signature_seal_test.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2026 Probo Inc <hello@probo.com>.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package coredata_test
+
+import (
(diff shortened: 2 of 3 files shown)


gearnode:
Merged thanks!

AnotherDork:
@gearnode Hijacking the PR comments as to not bloat into issues (If theres a better way to reach you, sorry. Happy to continue this elsewhere). Is there a specific reason for the 50k char/500k JSON bytes limit? Using the API generated path genuinely bypasses both: generated_document_service.go calls version.Insert() directly with no Validate() call anywhere in the file, and there's no DB-level constraint on the column. Same DocumentVersion type, same table, same storage. So it's a validation policy on one code path, not a storage or rendering constraint?

gearnode:
@AnotherDork those limits are arbitrary for now. We just want to make sure we have limits and input validation in place by default. We can adjust the values later, of course.

AnotherDork:
@gearnode Would it be considerable to raise that gap to 100k chars (+ json bytes proportionately). Admittedly asking for our company reasons, but doing field research across other company docset shows that the 50k limit is quite limiting for a well-rounded Policy. Since its arbitrary anyway it shouldn't cause any issues?

And separately, generated document validation bypass intended? (if not please dont change that :D)

gearnode:
@AnotherDork the latest release has increase this too 200k let us know if that suffisant for you. 

> And separately, generated document validation bypass intended? (if not please dont change that :D)

For now it's on purpose we'll keep this like that ;)

PR #1558, opened 2026-07-27
Cookie banner - Allow integrations in discovery mode

codenem:
## Summary by cubic
Enables discovery mode in `@probo/cookie-banner`: when no banner config is available, the client grants all consent types for supported integrations (Google Consent Mode) so GTM-managed tags can fire for detection. The deny-all bootstrap remains the default until a config is published.

### Package: cookie-banner **`+40`** **`-9`**
- Client: on config fetch failure, call `grantAll()` on integrations and start detectors.
- Integrations: add `grantAll()` to `ConsentIntegration`; implement for Google Consent Mode via a single `ALL_TYPES` list.
- GCM: keep bootstrap default as all denied; `grantAll()` updates all types to granted.
- Docs: add CHANGELOG entry for discovery-mode behavior.

<sup>Written for commit 85bdcffdbb070ef3b1fbd1367bc341508a20e849. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1715, opened 2026-08-17
Add option to disable Google Consent Mode in the cookie banner SDK

AnotherDork:
## Problem

The Google Consent Mode integration in `@probo/cookie-banner` is created unconditionally. For sites that already manage Consent Mode themselves (an existing CMP or GTM setup), this makes the SDK impossible to run alongside without consent side effects:

- `load()` fires `gtag('consent','default', ...)` denying **all seven** consent types before the config fetch — including types like `security_storage` that many sites deliberately leave unset (which Google treats as granted).
- When the config fetch fails, discovery mode fires `consent update` with everything **granted** (#1558), so an outage flips the site's consent signal.
- Both writes race the site's own `default`/`update` calls on `dataLayer`.

This matters in particular for self-hosted operators who want the tracker detection/reporting side of the SDK while keeping an existing, already-audited consent implementation.

A server-side toggle (like `capabilities.resourceReporting` from 0.259.0) can't solve this: `bootstrap()` runs **before** the config fetch, so no config field can suppress the initial deny-all. It has to be a client-side option.

## Change

- `CookieBannerClientOptions.integrations?: IntegrationConfig[]` — per-integration config array (currently one: `{ name: "gcm", enabled: boolean }`), default behavior unchanged for existing users.
- When GCM is disabled, `createDefaultIntegrations()` returns no integrations, so the SDK never touches `gtag`/`dataLayer`: no eager deny-all, no discovery-mode grant-all, no defaults or updates. Detection, consent recording, and resource activation are unaffected.
- Exposed as `gcm-enabled="false"` on `<probo-cookie-banner>` / `<probo-cookie-banner-root>` and `data-gcm-enabled="false"` on the IIFE script tag. The attribute accepts `"true"`/`"false"`; anything else logs a warning and falls back to enabled.
- Attribute parsing lives in a pure `resolveGcmEnabled()` predicate (mirroring `shouldStartResourceDetector`), unit-tested alongside `createDefaultIntegrations()`.
- README: new "Google Consent Mode" section documenting the switch. While editing the integrations bullet I also dropped its stale PostHog mention (integration removed in SDK 0.6.0).

Left `CHANGELOG.md` untouched per the release-time changelog practice; happy to add an Unreleased entry if preferred.

## Testing

- `npm --workspace @probo/cookie-banner run check` / `test` / `build` all pass (9 tests).
- `gcm-enabled` wiring verified present in the built `dist/cookie-banner.iife.js`.
- `lint-js` steps (`@probo/skills` validate, `make relay`, root eslint + n8n lint) run locally and pass.

### DIFF
--- a/packages/cookie-banner/src/integrations/index.ts
+++ b/packages/cookie-banner/src/integrations/index.ts
@@ -38,8 +38,9 @@ export function createDefaultIntegrations(
 
 // resolveGoogleConsentMode maps the `google-consent-mode` element attribute
 // (and the `data-google-consent-mode` script attribute) to the client option.
-// Only "off" and "false" disable the integration; absent or any other value
-// keeps the default behavior.
+// Only "off" and "false" (case-insensitive, whitespace-tolerant) disable the
+// integration; absent or any other value keeps the default behavior.
 export function resolveGoogleConsentMode(value: string | null): boolean {
-  return value !== "off" && value !== "false";
+  const normalized = value?.trim().toLowerCase();
+  return normalized !== "off" && normalized !== "false";
 }
--- a/packages/cookie-banner/src/integrations/index.ts
+++ b/packages/cookie-banner/src/integrations/index.ts
@@ -21,13 +21,15 @@
 export type { ConsentIntegration } from "./integration";
 export { GoogleConsentModeIntegration } from "./gcm";
 
+import type { IntegrationConfig } from "../types";
 import type { ConsentIntegration } from "./integration";
 import { GoogleConsentModeIntegration } from "./gcm";
 
 export function createDefaultIntegrations(
-  options?: { googleConsentMode?: boolean },
+  configs?: IntegrationConfig[],
 ): ConsentIntegration[] {
-  if (options?.googleConsentMode === false) {
+  const gcm = configs?.find((config) => config.name === "gcm");
+  if (gcm?.enabled === false) {
     return [];
   }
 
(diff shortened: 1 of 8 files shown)


packages/cookie-banner/src/types.ts
```diff
@@ -117,8 +117,14 @@ export interface ConsentRecord {
...
+// CookieBannerClientOptions configures the client. `googleConsentMode`
+// (default true) controls the built-in Google Consent Mode integration;
+// set it to false to skip it entirely — no eager deny-all default, no
+// discovery-mode grant-all, no consent defaults or updates — for sites
+// that manage Consent Mode themselves.
 export interface CookieBannerClientOptions {
   bannerId: string;
   baseUrl: string;
   lang?: string;
+  googleConsentMode?: boolean;
```
codenem: Let's have an integrations key here, would be an array of supported integrations config. For now we can have just one IntegrationConfig with an enabled field inside. This way if things evolve we have a scalable structure.

packages/cookie-banner/src/types.ts
```diff
@@ -117,8 +117,14 @@ export interface ConsentRecord {
   created_at: string;
 }
 
+// CookieBannerClientOptions configures the client. `googleConsentMode`
```
codenem: Let's strip those comments

packages/cookie-banner/src/integrations/index.ts
```diff
@@ -24,8 +24,23 @@ export { GoogleConsentModeIntegration } from "./gcm";
...
   return [
     new GoogleConsentModeIntegration(),
   ];
 }
+
+// resolveGoogleConsentMode maps the `google-consent-mode` element attribute
+// (and the `data-google-consent-mode` script attribute) to the client option.
+// Only "off" and "false" (case-insensitive, whitespace-tolerant) disable the
+// integration; absent or any other value keeps the default behavior.
+export function resolveGoogleConsentMode(value: string | null): boolean {
```
codenem: let's go for `gcm-enabled` for the attr name. So we can later prefix any other config field for gcm with `gcm-`. Let's allow `"true"/"false"` and log a warning when the value is not a valid one, and fallback to the default then (true).

codenem (changes requested):
Thanks for this! Left some comments.

AnotherDork:
applied in ff8ec26:

- `integrations?: IntegrationConfig[]` on the client options (one entry for now: `{ name: "gcm", enabled }`), replacing the boolean.
- Attribute renamed to `gcm-enabled` / `data-gcm-enabled`; accepts `"true"`/`"false"`, logs a warning and falls back to enabled on anything else.
- Comments on the options type stripped.

PR description updated to match.

codenem (approved):

codenem:
Thanks! Will rebase, merge and release today

PR #1732, opened 2026-08-18
Add GitHub App connector support

gearnode:
## Summary

- add GitHub App installation and callback support alongside GitHub OAuth and personal access tokens
- authorize each selected installation through a GitHub App user OAuth code before binding it to a Probo organization
- mint and cache short-lived installation access tokens while resolving app credentials from deployment config at use time (key rotation does not require reconnecting tenant rows)
- allow connector coexistence by protocol and use capability interfaces for organization selection, scope checks, reconnect, and probes
- expose typed `ConnectorProtocol` values and a provider-agnostic `configuredProtocols` catalog field
- derive console connect actions via Relay Resolver fields (`oauthConfigured`, `installableProtocols`) on `ConnectorProviderInfo`
- serve GitHub App installation callbacks on `/connectors/github-app/complete`, separate from OAuth2 `/connectors/complete`
- add bootstrap, environment, Helm, migration, and localization support

## Testing

- `go test ./pkg/connector/... ./pkg/accessreview/... ./pkg/probodconfig ./pkg/bootstrap ./pkg/probod ./pkg/server/api/console/v1/... -count=1`
- `golangci-lint run ./pkg/connector/... ./pkg/accessreview/... ./pkg/probodconfig/... ./pkg/bootstrap/... ./pkg/probod/... ./pkg/server/api/console/v1/...`
- `make relay`
- `npm run check --workspace=@probo/console`
- rendered the Helm chart with a GitHub App connector and verified both secrets and all `PROBOD_CONNECTOR_GITHUB_APP_*` environment variables

## Configuration

Enable **Request user authorization (OAuth) during installation** in the GitHub App settings and configure its callback URL as `/api/console/v1/connectors/github-app/complete`. Provide `PROBOD_CONNECTOR_GITHUB_APP_ID`, `PROBOD_CONNECTOR_GITHUB_APP_CLIENT_ID`, `PROBOD_CONNECTOR_GITHUB_APP_CLIENT_SECRET`, `PROBOD_CONNECTOR_GITHUB_APP_SLUG`, and `PROBOD_CONNECTOR_GITHUB_APP_PRIVATE_KEY`. OAuth2 connector redirect URIs stay on `/api/console/v1/connectors/complete`.

<div><a href="https://cursor.com/agents/bc-426a4862-b745-4133-83af-183dacb76383?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-426a4862-b745-4133-83af-183dacb76383&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---
## Summary by cubic
Adds a GitHub App connector alongside OAuth2 and API keys. It authorizes installations before binding, mints and caches short‑lived installation tokens, and makes probes and reconnects protocol‑aware; the console now derives available actions from a provider’s configured protocols. Also adds OAuth2 and SSH as access‑review auth methods and widens the DB unique key to include protocol.

### Tests **`+1842`** **`-1`**
Add e2e for GitHub App initiate/complete and configuredProtocols; unit tests for capability interfaces, protocol registry/lookup, completion routing, GitHub probe URL selection, builder/env wiring, and connector config round‑trips; expand GitHub driver tests (credentials, audit log, deploy keys).

### Coredata **`+62`** **`-4`**
Add ConnectorProtocol GITHUB_APP and include protocol in the (organization_id, provider, protocol) uniqueness constraint; add migrations. Extend AccessReviewEntryAuthMethod with OAUTH2 and SSH.

### GraphQL API **`+803`** **`-361`**
Expose connector.protocol, configuredProtocols, and canReconnect; serve GitHub App initiate and complete on dedicated paths; route completion via a shared state payload; gate organization pickers to OAuth2; add ConnectorProtocol enum and new auth methods to the schema; trim resolver special‑cases.

### MCP **`+2`** **`-0`**
Add OAUTH2 and SSH to the access‑review auth method enum.

### Service **`+2930`** **`-84`**
Implement the GitHub App connector: signed initiate state, installation authorization via user code, installation token minting with cache, and runtime injection of app credentials; support UnmarshalConnection for GITHUB_APP. Add capability interfaces (org selection, scope grant checks, reconnect, probe URL); make probes protocol‑aware and enforce reconnect protocol matches. Extend GitHub driver to collect OAuth apps, PATs, installations, and deploy keys, use audit log for last activity, and raise the source fetch timeout to 2 minutes. Register GitHub App from env/Helm at bootstrap/probod and add probodconfig support.

### App: console **`+210`** **`-65`**
Show a “GitHub App” action when installableProtocols includes GITHUB_APP; pass protocol on initiate/reconnect via a shared URL builder; derive oauthConfigured/installableProtocols from configuredProtocols via Relay resolvers; add OAuth2/SSH labels and responsive layout tweaks; update locales.

### Package: skills **`+1`** **`-1`**
Document OAuth2 and SSH in access‑review auth methods.

### prb (CLI) **`+2`** **`-2`**
Allow filtering by OAUTH2 and SSH auth methods.

### Other **`+57`** **`-1`**
Add .env guidance and Helm values/secrets for GitHub App (ID, slug, client ID/secret, private key); wire deployment env and secrets.
- Required: apply DB migrations; set PROBOD_CONNECTOR_GITHUB_APP_ID/CLIENT_ID/CLIENT_SECRET/SLUG/PRIVATE_KEY; configure the app callback as /api/console/v1/connectors/github-app/complete; update Helm values and secrets.

<sup>Written for commit a3e23b5fa71b55b639f1d4d83bd995ac4aaf4c09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1732?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/connector/github_app.go
```diff
@@ -0,0 +1,458 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot validate github app state token: %w", err)
+	}
+
+	setupAction := r.URL.Query().Get("setup_action")
+	if setupAction != "install" && setupAction != "update" {
+		return nil, nil, fmt.Errorf("cannot complete github app installation: invalid setup action")
+	}
+
+	installationID, err := strconv.ParseInt(r.URL.Query().Get("installation_id"), 10, 64)
```
aureliensibiril: `installation_id` comes straight off the query string and nothing ties it to the state. `fetchInstallationOrganization` authenticates with the app JWT, which reads any installation of our app, so the only checks left are "exists" and "targets an org".

Any org admin can start a real install to get a valid state, then replay the callback with someone else's `installation_id` (sequential ints, and GitHub documents these params as spoofable). Their installation binds into the attacker's org and the driver pulls the victim's members, roles and 2FA status.

Needs user authorization during install: exchange the `code`, confirm the installation is in `GET /user/installations` before persisting.

contrib/helm/charts/probo/templates/secret.yaml
```diff
@@ -71,6 +71,9 @@ stringData:
   {{- if .config.endpointToken }}
   connector-slack-endpoint-token: {{ .config.endpointToken | quote }}
   {{- end }}
+  {{- if and (eq .name "github") (eq .type "github-app") }}
+  connector-github-app-private-key: {{ .config.privateKey | quote }}
```
aureliensibiril, line 82: This sits inside `{{- if eq .name "slack" }}` (l.61), the three `end`s close github-if, slack-if, range. `.name` can't be both, so the key never renders, while `deployment.yaml:562` references it at range level with no `optional`. Configure the connector the way `values.yaml` documents it and every probod pod goes `CreateContainerConfigError`. Move it out one level.

pkg/connector/capabilities.go
```diff
@@ -0,0 +1,169 @@
...
+	apiBase, defaultProbe string,
+) (string, error) {
+	return ResolveProbeURL(connectionOrProbe(conn, protocol), apiBase, defaultProbe)
+}
+
+func (c *OAuth2Connection) SupportsOrganizationPicker() bool { return true }
+func (c *OAuth2Connection) SupportsScopeGrantCheck() bool    { return true }
+func (c *OAuth2Connection) SupportsReconnect() bool          { return true }
+
+func (c *APIKeyConnection) SupportsOrganizationPicker() bool { return false }
```
aureliensibiril: This breaks Cloudflare. `ProviderSupportsOrganizationPicker` was `cfg.ListOrgs != nil`, now it also needs `OrganizationSelector`, so it's false for every `API_KEY` connector. Cloudflare is API-key only, registers `ListOrgs`, and hard-fails on an empty `AccountID` with no `ExtraSetting` to collect one.

Auto-select bails, `SourceNeedsConfiguration` reports the source as configured, the picker returns nil, and every sync dies on `account_id is required` with no fix available from the UI. Gate on the provider config rather than the auth model, or keep `false` only on `GitHubAppConnection`.

pkg/connector/github_app.go
```diff
@@ -0,0 +1,458 @@
...
+		&struct {
+			Type           string `json:"type"`
+			AppID          string `json:"app_id"`
+			PrivateKey     string `json:"private_key"`
+			InstallationID int64  `json:"installation_id"`
+			APIBase        string `json:"api_base"`
+		}{
+			Type:           string(ProtocolGitHubApp),
+			AppID:          c.AppID,
+			PrivateKey:     c.PrivateKey,
```
aureliensibiril: The app private key ends up serialized in every connector row (copied at l.205, read back by `source_service` when it builds the client). Rotate it and every existing connector keeps signing with the revoked key until each one is reconnected by hand.

`ApplyManagedAPIKey` in that same function already does the opposite for managed keys. Persist `installation_id` and `api_base` only, inject app id and key from the registry.

Same PEM is also the `statelesstoken` HMAC secret, so a rotation kills in-flight states too.

pkg/server/api/console/v1/graphql/connector.graphql
```diff
@@ -166,6 +172,7 @@ input ConnectorFilter {
 type Connector {
     id: ID!
     provider: ConnectorProvider!
+    protocol: String!
```
aureliensibiril: Why `String` and not `coredata.ConnectorProtocol`? It exists, has a PG enum behind it, and `provider` right above uses `ConnectorProvider`. As written the frontend literals get no typecheck: an unknown protocol falls through `canReconnectConnector`'s `return true` and renders as `protocol.replaceAll("_", " ")`, skipping i18n in all three locales.

aureliensibiril (commented):
Went over the branch. Five things below, the first one is a blocker.

PR #1735, opened 2026-08-19
Allow multiple access-review connections per provider

aureliensibiril:
An organization can now connect the same access-review provider several times — two GitHub organizations, two Slack workspaces, two Render accounts — each connection backing its own access-review source. On top of that, every connector now has exactly **one owner**: an access-review source or a SCIM bridge, never both — which lets each owner delete its credential directly and removes the reference-counting row locks entirely.

## What changed

- **Schema**: the unique index on `connectors(organization_id, provider, protocol)` is dropped outright — no provider keeps a cap. A plain lookup index preserves the per-organization access path. Connector inserts can no longer conflict, so the typed already-exists mapping and the API conflict branches go away with it. Schema-forward-only (documented in the migration).
- **Single ownership**: partial unique indexes on `access_review_sources(connector_id)` and `iam_scim_bridges(connector_id)` make one-owner-per-table a schema invariant; cross-feature exclusivity is checked at the three bind sites by looking at the other referrer (a source refuses a bridge-held connector and vice versa). The checks are unlocked reads: a cross-bind would need two concurrent crafted API calls on the same fresh connector id — organic flows only ever bind their own freshly created connector — so the race is accepted rather than serialized (commented in code). The migration splits legacy shared rows by copying the connector (AES-GCM ciphertext is not bound to row identity).
- **Locks deleted**: `Connector.LockForUpdate` and `DeleteIfUnreferenced` (lock → count sources → count bridges → delete-if-zero) are gone. Source creation stays idempotent through the partial unique index (`INSERT … ON CONFLICT DO NOTHING`); `deleteAccessReviewSource` deletes source + connector in one transaction, reading the connector id from the DELETE's `RETURNING` so a concurrent relink cannot swap it unobserved; deleting a SCIM configuration locks the configuration row first (a concurrent bridge insert blocks on it through its FK), then deletes bridge and connector in order.
- **Slack messaging**: the **legacy messaging fallback** was the one consumer relying on per-provider uniqueness. Probot's slackbot path delivers via its own installation tokens and never reads the connectors table, so it is unaffected. The fallback now picks **deterministically**: a channel-configured connector wins over newer plain rows, tie-broken by oldest. Cross-workspace message routing remains future product work.
- **Explicit reconnect**: a bare `/connectors/initiate` always creates a new connector; reconnect happens only via the explicit `connector_id` parameter. The implicit `(org, provider)` fallback, its service getter, and the generic by-provider loaders are deleted.
- **Idempotent source creation**: `createAccessReviewSource` with a `connectorId` returns the existing source (`created: false`) when one already references the connector — on **GraphQL and MCP** alike, arbitrated by the unique index instead of a row lock. Two tabs replaying a callback cannot double-create.
- **Atomic SCIM setup**: `createSCIMConfiguration` creates the configuration and its optional bridge in one transaction — a refused connector (source-held or wrong provider) can no longer commit a bridgeless configuration that would block every retry (configurations are unique per organization). The service-only `DeleteSCIMBridge` (bridge-without-connector leak, no callers) is deleted — no API surface change.
- **Console**: the catalog never hides a connected provider ("Available" section); a Relay updater inserts edges only for genuinely new sources; provider-denied OAuth errors are toasted instead of latching in the URL.
- **Observability**: fetch-failure and resolved-name logs are keyed by connector GID; resolved instance names no longer appear in log fields (logging.md rule).

## Behavior notes (reviewed, documented in code)

- **Relinking or unlinking a source now deletes the abandoned connector** — previously it was stranded forever. The credential dies with its last owner, in the same transaction.
- **`deleteConnector` on a connector held by a live SCIM bridge now fails** (FK flipped from `SET NULL` to plain) instead of silently disabling sync.
- A legacy org whose only Slack integration is the old channel-configured messaging connector and clicks the SLACK catalog tile gets a **second** Slack connector + source instead of having its messaging credential overwritten — strictly safer; the messaging pick is pinned by a DB test.
- `slackConnections` (compliance-portal integrations page) lists every Slack connector including access-review ones — read-only listing.
- An abandoned flow (tab closed between callback and source mutation) strands an unreferenced connector; fast-follow: an orphan-connector sweep worker.
- The pre-existing OAuth state replay/session-binding gap is out of scope; recommended as the immediate companion PR.

## Testing

- DB tests against real Postgres (all migrations applied): several same-provider connectors insert; the idempotent-insert conflict path; the bridge unique index (23505 pinned by constraint name); the Slack messaging selector pick. The migration's dedup was additionally validated against seeded legacy data (worst case: one connector with 3 sources + a bridge → 4 single-owner connectors, snapshots intact, zero sharing left).
- e2e: two connectors + two sources of one provider coexist; repeat creation returns the existing source (`created: false`); deleting a source deletes exactly its own connector (survivor asserted by ID); **relinking a source deletes the abandoned connector**.
- Multiple adversarial review rounds (internal, gearnode-lens, OpenAI Codex) on plan and code; every commit builds independently (bisect-safe).


pkg/coredata/connector.go
```diff
@@ -340,6 +324,84 @@ WHERE %s AND id = @id
...
+		if errors.Is(err, ErrResourceNotFound) {
+			return nil
+		}
+
+		return fmt.Errorf("cannot lock connector: %w", err)
+	}
+
+	sources := &AccessReviewSources{}
+
+	sourceCount, err := sources.CountByConnectorID(ctx, conn, scope, connectorID)
```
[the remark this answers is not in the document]
aureliensibiril: Pre-existing, deliberate semantics rather than a regression: this helper serializes exactly the reference check DeleteSource already performed (live sources + SCIM bridges). `access_review_campaign_sources.connector_id` deliberately has no FK to `connectors` — snapshots are append-only copies designed to survive connector deletion (see 20260611T010000Z.sql). When a source is deleted mid-campaign, the fetch path already handles the missing credential as a failed fetch attempt ("source fetch failed but campaign can continue") and the campaign lifecycle finalizes. Keeping the connector alive until campaigns finish would change that designed behavior and is out of scope here.

pkg/connector/oauth2.go
```diff
@@ -116,11 +116,16 @@ type (
...
-		RequestedScopes []string `json:"scopes,omitempty"`
+		OrganizationID string `json:"oid"`
+		Provider       string `json:"provider"`
+		ContinueURL    string `json:"continue,omitempty"`
+		ConnectorID    string `json:"cid,omitempty"` // Set when reconnecting an existing connector
+		// ExplicitReconnect distinguishes a reconnect the caller selected
+		// with an explicit connector_id from the initiate handler's
+		// (org, provider) fallback for single-connection providers. The
+		// callback redirect's flow param depends on it.
+		ExplicitReconnect bool     `json:"xr,omitempty"`
```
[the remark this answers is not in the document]
aureliensibiril: Known and accepted version-skew window, documented in the PR description and the frontend comment: a pre-change state completing on a new binary deserializes `xr` as false and redirects `flow=create`, but the console's create path keeps the `hasSourceForCallback` dedup guard precisely for redirects minted without the contract — a reconnect of an existing source skips creation. The residual exposure (source beyond the first 50 loaded edges) is bounded by the 10-minute state TTL times the deploy overlap and produces a visible, deletable duplicate source; server-side source idempotency is the planned follow-up.

pkg/connector/provider/slack.go
```diff
@@ -33,6 +33,10 @@ func slackRegistration() *Registration {
 	return &Registration{
 		Provider:    coredata.ConnectorProviderSlack,
 		DisplayName: "Slack",
+		// The Slack messaging worker (pkg/slack/worker.go) resolves the
+		// organization's credential by provider alone, so at most one
+		// Slack connector may exist per organization.
+		SingleConnection: true,
```
gearnode: no required anymore as the bot has it's own config now.

aureliensibiril: Agreed, and the current revision goes exactly there: the SingleConnection trait, the SLACK partial unique index, and the initiate fallback are all gone (this hunk no longer exists). No provider keeps a cap. The only Slack-specific piece left is LoadSlackMessagingConnector, which protects whatever traffic still routes through probot's queueLegacy fallback (orgs with no slackbot destination): it picks the channel-configured row first, then oldest, so adding access-review workspaces never re-routes legacy messaging. If the legacy fallback itself is due for deletion, that selector goes with it.

pkg/connector/oauth2.go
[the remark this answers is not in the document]
aureliensibiril: Resolved by the latest revision: the ExplicitReconnect bit and the flow param no longer exist. Source creation is now server-authoritatively idempotent per connector (FOR UPDATE row lock + created flag on the payload), so version-skewed states cannot misclassify anything — a replayed or old-state callback simply resolves to the existing source.

apps/console/src/pages/organizations/access-reviews/dialogs/accessReviewSourceMutations.ts
```diff
@@ -18,22 +18,49 @@
...
+  if (!edge || !node || !connection) return;
+
+  const nodeId = node.getDataID();
+  const edges = connection.getLinkedRecords("edges") ?? [];
+  if (edges.some(e => e?.getLinkedRecord("node")?.getDataID() === nodeId)) {
+    return;
+  }
+
+  ConnectionHandler.insertEdgeBefore(connection, edge);
+}
```
gearnode, line 66: we should use relay directive.

pkg/accessreview/source_service.go
```diff
@@ -95,13 +95,18 @@ func (r *UpdateAccessReviewSourceRequest) Validate() error {
...
+// connector: when a source already references req.ConnectorID, it is
+// returned with created=false instead of inserting a duplicate. The
+// connector row lock serializes concurrent callers so exactly one
+// creates.
 func (s *Service) CreateSource(
 	ctx context.Context,
 	scope coredata.Scoper,
 	req CreateAccessReviewSourceRequest,
-) (*coredata.AccessReviewSource, error) {
+) (*coredata.AccessReviewSource, bool, error) {
```
gearnode: s/CreateSource/UpsertSource/ ?

pkg/accessreview/source_service.go
```diff
@@ -187,9 +215,28 @@ func (s *Service) UpdateSource(
...
+						return fmt.Errorf("cannot serialize source relink: %w", err)
+					}
+
 					if err := connector.LoadMetadataByID(ctx, conn, scope, **req.ConnectorID); err != nil {
 						return fmt.Errorf("cannot load connector: %w", err)
 					}
+
+					if connector.OrganizationID != source.OrganizationID {
+						return fmt.Errorf("cannot update access source: organization mismatch")
+					}
```
gearnode: this check should be useless.

pkg/accessreview/source_service.go
```diff
@@ -187,9 +215,28 @@ func (s *Service) UpdateSource(
 			if req.ConnectorID != nil {
 				if *req.ConnectorID != nil {
 					connector := &coredata.Connector{}
+					if err := connector.LockForUpdate(ctx, conn, scope, **req.ConnectorID); err != nil {
+						return fmt.Errorf("cannot serialize source relink: %w", err)
+					}
+
```
gearnode: Why lock it?

pkg/coredata/access_review_source.go
```diff
@@ -308,6 +308,59 @@ WHERE %s AND id = @id
 	return nil
 }
 
+// LoadByConnectorID loads the access source referencing the connector,
+// oldest first when several exist (possible only in pre-existing data:
+// creation is idempotent per connector on every surface). Returns
+// ErrResourceNotFound when none references it.
```
gearnode: drop comment.

pkg/coredata/connector.go
```diff
@@ -340,6 +333,83 @@ WHERE %s AND id = @id
...
+	conn pg.Tx,
+	scope Scoper,
+	connectorID gid.GID,
+) error {
+	if err := c.LockForUpdate(ctx, conn, scope, connectorID); err != nil {
+		if errors.Is(err, ErrResourceNotFound) {
+			return nil
+		}
+
+		return fmt.Errorf("cannot serialize connector deletion: %w", err)
```
gearnode: we do we lock here?

pkg/coredata/connector.go
```diff
@@ -340,6 +333,83 @@ WHERE %s AND id = @id
...
+	if err != nil {
+		return fmt.Errorf("cannot count scim bridges for connector: %w", err)
+	}
+
+	if bridgeCount > 0 {
+		return nil
+	}
+
+	return c.Delete(ctx, conn, scope)
+}
```
gearnode: since we now can have many connectors i would have a dedicated one for SCIM this will remove this code path.

pkg/accessreview/source_service.go
aureliensibiril: Done — renamed to `EnsureSource`.

pkg/accessreview/source_service.go
aureliensibiril: Dropped.

pkg/accessreview/source_service.go
aureliensibiril: Lock removed — the partial unique index on connector_id arbitrates concurrent creates now.

pkg/coredata/access_review_source.go
aureliensibiril: Dropped.

pkg/coredata/connector.go
aureliensibiril: Gone — DeleteIfUnreferenced and the lock are deleted; each owner deletes its connector directly in its own transaction.

pkg/coredata/connector.go
aureliensibiril: Done via referrer ownership rather than a dedicated connector kind: a source refuses a bridge-held connector and a bridge refuses a source-held one (partial unique indexes underneath), and this code path is deleted.

PR #1737, opened 2026-08-19
Add a per-organization cloud identity issuer

lukkor:
## Summary

Phase 1 of the AWS OIDC audit connector: probod becomes an outbound OIDC
identity provider so customers can grant read-only audit access to their cloud
accounts without sharing any secret with Probo.

- `pkg/cloudidentity` mints short-lived RS256 tokens (`sub =
  organization:<orgGID>`, 5-minute TTL, seven claims and nothing else) and
  builds the discovery and JWKS documents. Tokens are minted in-process; there
  is deliberately no token endpoint.
- `pkg/server/cloudidentity` serves the two public documents at
  `GET /identity/org/{orgGID}/.well-known/openid-configuration` and
  `GET /identity/org/{orgGID}/jwks`, both with `Cache-Control: public,
  max-age=3600`. Only the appended OIDC Discovery form exists, since AWS and GCP
  follow it; the RFC 8414 variant the OAuth2 server serves has no equivalent
  here.
- `issuer` and `jwks_uri` always come from configuration, never from the request
  `Host`, so an arbitrary Host cannot obtain a document claiming to be an issuer
  it controls.
- New `cloud-identity` config section (`enabled`, `issuer-base-url`,
  `signing-keys`) with `PROBOD_CLOUD_IDENTITY_*` env vars, propagated through
  every file in the config checklist. Its signing keys are separate from the
  OAuth2 server's.
- `Organization.ExistsByID` is an unscoped existence check, so a mistyped
  identifier 404s when the customer registers the provider instead of
  registering one that can never work. It is kept to an existence check so no
  organization data is reachable without a scope.

**Off by default.** Existing self-hosted and Helm deployments keep starting with
no second RSA key and serve no `/identity` route. SaaS turns it on with
`https://proboidentity.com`.

## Reviewer notes

- **The advertised issuer URL is immutable once customers register it** with
  their cloud provider; changing it later forces every customer to redeploy
  infrastructure they own. `https://proboidentity.com` is the value that becomes
  permanent.
- The DNS and CloudFront work for that apex is Phase 0 and is **not** in this
  PR. Nothing breaks without it, because no customer registers an issuer until
  the Phase 4 templates exist.
- Startup applies the strict AWS URL rules (https, no port, no query, issuer
  under 255 characters) only to an explicitly configured non-loopback base. A
  base derived as `{base-url}/identity` is exempt, otherwise a self-hosted
  deployment on http or a port could not start. The collision rule, length
  ceiling and active-key requirement always apply.
- `WORKLOAD_IDENTITY`, the AWS provider, CloudFormation and Terraform artifacts,
  the discovery worker, and the GraphQL/MCP/CLI/UI surface are phases 2 to 6 and
  out of scope.

## Test plan

- [x] `go test ./pkg/cloudidentity/...` — claim set and TTL, subject format,
      issuer and JWKS URL construction for both the SaaS apex and the derived
      self-hosted base, issuer-URL validation table, and a self-verification
      round trip that mints a token and verifies it against the served JWKS
      (this is the class of kid/JWKS wiring bug that surfaces in production as
      an opaque `AccessDenied`)
- [x] Golden file pins the discovery document, so a refactor cannot quietly drop
      one of the five fields AWS requires
- [x] Key rotation: a retired but still published key verifies; a key dropped
      from configuration does not; inactive keys never sign
- [x] `go test ./pkg/server/cloudidentity/...` — unknown organization, wrong
      entity type, mis-cased and malformed identifiers all 404; documents are
      unaffected by a spoofed request `Host`; JWKS never reads the database; no
      token endpoint and no write methods
- [x] `go test ./pkg/bootstrap/...` — the signing key is required only when
      enabled
- [x] `make test-e2e` (`TestCloudIdentity_*`) — discovery and JWKS fetched over
      HTTP from a real probod subprocess, unknown organization 404s
- [x] Manually started probod with the issuer disabled: it boots with no
      signing key and `/identity/...` falls through to the console SPA
- [x] Manually started probod with the issuer enabled: it logs the effective
      issuer base and no key material
- [x] `make go-fmt`, `go vet`, `go build ./...` clean
- [x] `helm template` against the chart — could not run locally (`helm` is not
      installed); the three template edits copy the existing `slackbot` gating
      pattern and were reviewed by hand

---
## Summary by cubic
Adds a per-organization outbound OIDC issuer so customers can grant Probo cloud access without shared secrets. Also decodes private keys at config load and switches OAuth2 signing to a shared `pkg/crypto/jose.KeyRing`, so key errors fail fast, rotation is safer, and base URL parsing is stricter; the issuer is off by default.

### Tests **`+2169`** **`-32`**
End-to-end tests cover discovery and JWKS at the derived issuer base. Unit tests cover issuer config/TTL/rotation, RSA public key validation and EC JWK parsing, handler hardening and org validation, base URL parsing (empty host/port), bootstrap key requirements, and a golden discovery document.

### GraphQL API **`+246`** **`-30`**
Adds `pkg/server/identityfederation` handlers serving per-org `/.well-known/openid-configuration` and `/jwks` under the issuer base, cacheable for 1h. Validates org GID type/casing and existence, ignores request Host, and only mounts when the issuer is enabled. Extracts HTTP cache helpers to `pkg/server/httpx` and updates OAuth2 handlers to use them.

### Service **`+1283`** **`-231`**
Introduces `pkg/identityfederation` (claims, config, issuer) to build per-org discovery/JWKS and mint 5‑minute RS256 tokens (no token endpoint), with keys separate from OAuth2. Adds `pkg/crypto/jose.KeyRing` and RSA public key validation; `pkg/iam/oauth2` now signs via the KeyRing (drops local key round‑robin). Fixes EC JWK parsing in `pkg/iam/oidc`. Hardens `pkg/baseurl` (reject empty host/empty port). Decodes private keys at config load via `pkg/probodconfig.PrivateKey`/`RSAPrivateKey` (applies to OAuth2, identity federation, SAML, and ACME) and updates `pkg/bootstrap` to read `PROBOD_IDENTITY_FEDERATION_*`, construct signing keys, and parse the ACME account key. Adds `OrganizationService.Exists` as a cross‑tenant existence check used by public discovery. Wires the new config and key types through `pkg/probod`.

### Agents **`+1`** **`-1`**
Updates the configuration checklist to include identity federation and the dev signing key.

### Other **`+138`** **`-3`**
Helm adds `probo.identityFederation.*` values and secret templates; `.env.example`, `make dev-config`, and the Lima sandbox generate and enable a separate identity‑federation RSA key.
- To enable: set `PROBOD_IDENTITY_FEDERATION_ENABLED=true` and provide a PEM key via `PROBOD_IDENTITY_FEDERATION_SIGNING_KEY` (do not reuse `oauth2.signingKey`).
- Optional: set `PROBOD_IDENTITY_FEDERATION_ISSUER_BASE_URL` to a public HTTPS apex; treat it as immutable after customers register it. Helm equivalents are under `probo.identityFederation.*` (secrets: `identity-federation-signing-key`, and optional previous key).

<sup>Written for commit d2f9f0f135313bd8e854f0df9f78cad318e13be0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1737?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lukkor:
@cubic-dev-ai review again please 🙏

pkg/baseurl/baseurl.go
```diff
@@ -53,10 +53,19 @@ func Parse(rawURL string) (*BaseURL, error) {
...
+	if parsed.Hostname() == "" {
 		return nil, fmt.Errorf("base URL must include a host")
 	}
 
+	// Port is also empty for a bare delimiter, so a trailing colon needs its
+	// own check.
+	if strings.HasSuffix(parsed.Host, ":") {
+		return nil, fmt.Errorf("base URL host has an empty port")
+	}
+
```
gearnode, line 67: let use https://pkg.go.dev/net#SplitHostPort ?

pkg/coredata/organization.go
```diff
@@ -99,6 +99,33 @@ func (o Organization) CursorKey(orderBy OrganizationOrderField) page.CursorKey {
...
+	args := pgx.StrictNamedArgs{"organization_id": organizationID}
+
+	var exists bool
+	if err := conn.QueryRow(ctx, q, args).Scan(&exists); err != nil {
+		return false, fmt.Errorf("cannot check organization existence: %w", err)
+	}
+
+	return exists, nil
+}
+
```
gearnode: Use already existing Load here.

pkg/crypto/jose/jose.go
```diff
@@ -110,6 +110,38 @@ func SignJWT(privateKey *rsa.PrivateKey, kid string, claims any) (string, error)
 	return signingInput + "." + signatureB64, nil
 }
 
+// ValidateRSAPublicKey reports whether pub is acceptable for RS256 JWK use.
+func ValidateRSAPublicKey(pub *rsa.PublicKey) error {
+	if pub == nil || pub.N == nil {
+		return fmt.Errorf("cannot validate rsa public key: public key is required")
+	}
```
gearnode, line 117: do we need defensive code like that?

pkg/crypto/jose/jose.go
```diff
@@ -110,6 +110,38 @@ func SignJWT(privateKey *rsa.PrivateKey, kid string, claims any) (string, error)
 	return signingInput + "." + signatureB64, nil
 }
 
+// ValidateRSAPublicKey reports whether pub is acceptable for RS256 JWK use.
+func ValidateRSAPublicKey(pub *rsa.PublicKey) error {
```
gearnode, line 114: Why do we need to expose this one publiclu?

pkg/iam/oidc/service.go
```diff
@@ -734,11 +734,17 @@ func parseJWK(k jwk) (crypto.PublicKey, error) {
...
+		uncompressed := make([]byte, 1+len(xBytes)+len(yBytes))
+		uncompressed[0] = 0x04
+		copy(uncompressed[1:], xBytes)
+		copy(uncompressed[1+len(xBytes):], yBytes)
+
+		pub, err := ecdsa.ParseUncompressedPublicKey(curve, uncompressed)
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse EC public key: %w", err)
+		}
+
```
gearnode: ?

lukkor: I'll remove the change, I did it because of the staticcheck of golangci-lint; I'll put a nolint instruction instead. But also, EC keys are not used by neither Google nor Microsoft (RSA only) so we may remove the EC branch completely in the future.

pkg/identityfederation/claims.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Metadata is the OIDC discovery document for one organization's issuer.
+	// AWS rejects an OIDC provider whose document omits any of these fields.
+	Metadata struct {
+		Issuer                           string   `json:"issuer"`
+		JWKSURI                          string   `json:"jwks_uri"`
+		ResponseTypesSupported           []string `json:"response_types_supported"`
+		SubjectTypesSupported            []string `json:"subject_types_supported"`
+		IDTokenSigningAlgValuesSupported []string `json:"id_token_signing_alg_values_supported"`
+	}
```
lukkor, line 49: Not shared with OAuth2 server/pkg as claims are really specific/minimal in this case and not used at all (in particular ndf and jti).

pkg/probo/organization_service.go
```diff
@@ -108,6 +108,26 @@ func (s OrganizationService) Get(
 	return organization, nil
 }
 
+// Exists reports whether the organization is present. The lookup is
+// deliberately cross-tenant: identity federation discovery is public, and the
+// caller has no identity to scope by. Only the boolean escapes, so no
+// organization data is reachable without a scope.
+func (s OrganizationService) Exists(
```
gearnode, line 115: have scope here.

pkg/server/identityfederation/handler.go
```diff
@@ -0,0 +1,117 @@
...
+		)
+		jsonx.RenderInternalServerError(w)
+
+		return
+	}
+
+	w.Header().Set(
+		"Cache-Control",
+		fmt.Sprintf("public, max-age=%d", int(documentCacheMaxAge.Seconds())),
+	)
```
gearnode: PublicCache

pkg/server/identityfederation/handler.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil || organizationID == gid.Nil || organizationID.EntityType() != coredata.OrganizationEntityType {
+		jsonx.RenderNotFound(w, fmt.Errorf("not found"))
+
+		return
+	}
+
+	w.Header().Set(
+		"Cache-Control",
+		fmt.Sprintf("public, max-age=%d", int(documentCacheMaxAge.Seconds())),
+	)
```
gearnode: PublicCache

PR #1740, opened 2026-08-20
Fall back to schema defaults for absent node attrs

AnotherDork:
## Problem

`prosemirror` node attribute accessors unmarshal `n.Attrs` unconditionally, so a node that omits `attrs` fails with `unexpected end of JSON input`. Attributes are optional in the Tiptap schema — the editor renders such a document fine by applying the schema defaults — but one attr-less table cell aborts `RenderHTML` for the whole document.

`docgen.ProseMirrorJSONToHTML` turns any render error into `<p>{escaped JSON}</p>`, so the failure surfaces as a PDF whose body is the raw ProseMirror JSON: document export, the employee signature screen, and the approval review screen all show JSON where the document should be. Nothing logs, and the same document looks correct in the editor.

## Reproduce

```go
n, err := prosemirror.Parse(`{"type":"tableCell","content":[{"type":"paragraph","content":[{"type":"text","text":"X"}]}]}`)
// err == nil

_, err = prosemirror.RenderHTML(n)
// cannot render td node: cannot parse table cell attrs: unexpected end of JSON input
```

Same shape for `{"type":"heading","content":[…]}` (renders as `invalid level 0`), `{"type":"orderedList","content":[…]}` and `{"type":"image"}`.

End to end: create a document whose content JSON omits `attrs` on table cells — `createDocument` accepts it and `SanitizeDocumentJSON` preserves it — open it in the editor (correct), then export the version as PDF or open it on the approval review screen: the body is the JSON dump.

## Fix

Apply the Tiptap schema defaults when `attrs` is absent — heading `level: 1`, ordered list `start: 1`, table cell/header `colspan`/`rowspan: 1` — mirroring the guard `CodeBlockAttrs` already had. Explicit values still win, and `"attrs":{}` with individual keys missing now gets the same defaults.

Verified against a real 13-page document that previously rendered as JSON; it now renders identically to the same document with full cell attrs.

## Note, not changed here

`ProseMirrorJSONToHTML` converting any render error into an escaped-JSON body means a rendering bug ships silently into an approval-grade, watermarked PDF that someone is asked to sign. Surfacing the error instead looks safer, but it changes export behaviour, so I left it out of this PR — happy to follow up if you want it.

---
## Summary by cubic
Fall back to Tiptap schema defaults when ProseMirror nodes omit attrs so documents render instead of failing. Previously, missing attrs caused unmarshal errors that surfaced as escaped JSON in PDFs and approval/signature screens; now `RenderHTML` renders those nodes using defaults.

### Tests **`+157`** **`-0`**
Add renderer and accessor coverage for nodes without attrs and with empty attrs objects: heading (level 1), ordered list (start 1), table cell/header (colspan/rowspan 1), image (empty src), and code block (nil language). Assert both HTML output and per-accessor defaults.

### Service **`+21`** **`-3`**
Return schema defaults when `n.Attrs` is absent or empty in `HeadingAttrs`, `OrderedListAttrs`, `ImageAttrs`, and `TableCellAttrs`, matching the existing `CodeBlockAttrs` guard; explicit values still win. Behavior change: heading level defaults to 1; ordered list start to 1; table cell/header colspan/rowspan to 1; empty `attrs: {}` treated the same.

<sup>Written for commit 53a2f104f67964d1d6420ba1678d989f15ab1dbb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1740?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/prosemirror/html_renderer_test.go
+++ b/pkg/prosemirror/html_renderer_test.go
@@ -418,3 +418,58 @@ func TestRenderHTML_UnknownMarkType(t *testing.T) {
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "unknown type")
 }
+
+func TestRenderHTML_NodesWithoutAttrs(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name     string
+		raw      string
+		expected string
+	}{
+		{
+			name:     "heading",
+			raw:      `{"type":"heading","content":[{"type":"text","text":"X"}]}`,
+			expected: `<h1>X</h1>`,
+		},
+		{
+			name:     "ordered list",
+			raw:      `{"type":"orderedList","content":[{"type":"listItem","content":[{"type":"paragraph","content":[{"type":"text","text":"item"}]}]}]}`,
+			expected: `<ol><li><p>item</p></li></ol>`,
+		},
+		{
+			name:     "table cell",
+			raw:      `{"type":"tableCell","content":[{"type":"paragraph","content":[{"type":"text","text":"X"}]}]}`,
+			expected: `<td><p>X</p></td>`,
+		},
+		{
+			name:     "table header",
+			raw:      `{"type":"tableHeader","content":[{"type":"paragraph","content":[{"type":"text","text":"X"}]}]}`,
+			expected: `<th><p>X</p></th>`,
+		},
+		{
+			name:     "empty attrs object",
+			raw:      `{"type":"tableCell","attrs":{},"content":[{"type":"paragraph","content":[{"type":"text","text":"X"}]}]}`,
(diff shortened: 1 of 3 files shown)


AnotherDork:
Happy to send you a screenshot of the issue to discord/DM. Didnt want to post snipets of our document to public PR :D

codenem:
Thanks @AnotherDork this is merged and will be released soon

PR #1766, opened 2026-08-24
Add a workload identity connector protocol

lukkor:
Probo can now reach a customer's cloud account while holding no credential for it. A new WORKLOAD_IDENTITY protocol mints a short-lived OIDC assertion through pkg/identityfederation, and the customer's STS exchanges it for temporary read-only credentials against a role they own and can revoke without coordinating with us.

The connection stores nothing at all, so it cannot leak a credential. Reaching that meant taking Client() off the Connection interface into a new HTTPConnection. A credential a cloud SDK uses to sign requests it builds itself has no transport to carry it, and faking a Client that always failed forced every caller to carry a sentinel error plus a capability probe to detect the lie. Callers now type-assert, and the four sites in pkg/accessreview that assumed HTTP branch explicitly — including the connection-status resolver, which mapped any credential error to DISCONNECTED and so would have shown a permanent red badge for a perfectly healthy cloud connector.

pkg/cloud holds the single shared abstraction, a Session meaning authenticated access to one account. pkg/cloud/aws vends credentials through stscreds with an in-process token retriever, so no assertion ever reaches disk. Its test drives a fake STS that really fetches the published JWKS, verifies the RS256 signature and evaluates the aud and sub conditions of a trust policy — including the case releases must gate on: one organization's identity cannot reach another's role.

Adding the protocol also exposed Registration as a god object, with 36 flat fields where a handful apply to any one provider and six startup rules whose only job was stopping two fields from disagreeing. Connect paths are now four nil-able blocks, so presence replaces the Supports* booleans and those six rules describe states that can no longer be written down. The four mutually exclusive API-key presentations became one mode enum, and Registry's four forwarding accessors collapsed into NewAPIKeyConnection, the single place a mode becomes wire format.


gearnode (approved):
LGTM

pkg/cloud/aws/session.go
```diff
@@ -0,0 +1,223 @@
...
+	opts ...Option,
+) (*Session, error) {
+	if issuer == nil {
+		return nil, fmt.Errorf("cannot open aws session: identity federation issuer is required")
+	}
+
+	if organizationID == gid.Nil {
+		return nil, fmt.Errorf("cannot open aws session: organization is required")
+	}
+
```
gearnode: ?

PR #1768, opened 2026-08-24
Add the AWS connector and its cloud accounts

lukkor:
AWS is the first provider on the WORKLOAD_IDENTITY protocol, and it declares only that connect path — no OAuth2 block, no API key, no client credentials — because there is no credential for a customer to paste or an operator to configure. The customer grants access in their own account, so the driver catalog admits it with no deployment configuration at all. Its connection check asks STS who we are, the one call every principal may make, so a failure means the credential exchange itself was refused rather than a permission being absent.

One connector covers a whole AWS organization: connectors are unique per (organization_id, provider), so the accounts beneath one need rows of their own. cloud_accounts is therefore keyed on the connector and the provider's own identifier, and names no cloud anywhere, so GCP projects land in the same table rather than motivating a parallel one. Its last_error_code holds a code rather than the provider's message, because AWS error text carries account IDs and ARNs.

AWSConnectorSettings records the issuer the connector was registered under. The customer pinned that exact URL in an OIDC provider and a trust policy that only they can change, so a deployment that moves its issuer strands every connector that exists. Recording it is what a gradual migration would later be built on; for now the session refuses a mismatch by name, since minting under a retired issuer also means serving its discovery and JWKS documents, which probod does not do.

The access review driver is not implemented and returns an error. Register refuses a workload identity path with a nil driver, so the alternative was leaving AWS out of the registry entirely; nothing reaches the stub either way, because no create path for the protocol exists yet.

Two registry tests read every provider as an HTTP one, requiring a probe URL and a non-nil NewDriver. Both now accept the cloud path — a gap since the protocol landed, which AWS is simply the first provider to expose.

---
## Summary by cubic
Adds AWS as the first WORKLOAD_IDENTITY connector. It uses assertion-based auth with no customer-provided secrets, records `AWSConnectorSettings` per connector, and marks AWS as coming soon in the console's access review source list.

### Coredata **`+51`** **`-1`**

- Registers `ConnectorProviderAWS` and adds `AWSConnectorSettings` with role ARN, pinned issuer, and a `DefaultAWSRoleName` constant.
- Adds a migration for the `AWS` enum value.

### Service **`+126`** **`-27`**

- Registers AWS under workload identity only, with no HTTP endpoint override and a cloud-based probe.
- Adds `Session.CheckAccess` using `sts:GetCallerIdentity`, and drops the STS endpoint and HTTP client override options.
- Stubs the access review driver, which returns an error until implemented.

### GraphQL API **`+1`** **`-0`**

- Adds `AWS` to the `ConnectorProvider` enum.

### App: console **`+21`** **`-7`**

- AWS rows show a "Coming soon!" badge; its connect actions are hidden because access review is not implemented.

### Package: ui **`+39`** **`-0`**

- Adds the AWS logo to `ThirdPartyLogo`.

### Tests **`+169`** **`-723`**

- Replaces the elaborate fake STS and its end-to-end exchange tests with a lighter session test.
- Covers the AWS registry shape and session construction, and updates registry and probe tests to accept the cloud path.

<sup>Written for commit 421933d226bebe43f93bcad2ccd2cb55ac9867e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1768?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/connector/provider/aws.go
```diff
@@ -0,0 +1,195 @@
...
+// An empty setting means the connector predates the field, so it is trusted to
+// be on the current issuer.
+func checkAWSIssuer(
+	issuer *identityfederation.Issuer,
+	organizationID gid.GID,
+	settings coredata.AWSConnectorSettings,
+) error {
+	if settings.Issuer == "" {
+		return nil
+	}
```
gearnode: drop

pkg/connector/provider/aws.go
```diff
@@ -0,0 +1,195 @@
...
+	if !awsRoleNamePattern.MatchString(roleName) {
+		return "", fmt.Errorf("cannot build aws role ARN: %q is not an IAM role name", roleName)
+	}
+
+	return arn.ARN{
+		Partition: awsPartition,
+		Service:   awsIAMService,
+		AccountID: settings.AccountID,
+		Resource:  "role/" + roleName,
+	}.String(), nil
```
gearnode: let only use roleARN has input.

pkg/connector/provider/aws.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot resolve aws connector issuer: %w", err)
+	}
+
+	if settings.Issuer != current {
+		return fmt.Errorf(
+			"cannot reach aws connector: it was registered under issuer %q, which this deployment no longer mints",
+			settings.Issuer,
+		)
+	}
```
gearnode: ??

pkg/coredata/entity_type_reg.go
```diff
@@ -158,7 +158,7 @@ const (
 	AgentExecutionAnchorEntityType                   uint16 = 126
 	BotMessageEntityType                             uint16 = 127
 	BotThreadSubjectEntityType                       uint16 = 128
-	TreatmentPlanEntityType                          uint16 = 129
+	_                                                uint16 = 129 // CloudAccountEntityType - never assigned
```
gearnode: drop

pkg/cloud/aws/session.go
```diff
@@ -61,6 +61,10 @@ type (
 	Session struct {
 		cfg       awssdk.Config
 		accountID string
+		// stsEndpoint is kept so that a client built after construction
+		// reaches the same STS the exchange did. Without it a test pointing
+		// the exchange at a fake would still send CheckAccess to real AWS.
+		stsEndpoint string
```
gearnode: if only for test let drop it

pkg/connector/provider/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	_ context.Context,
+	issuer *identityfederation.Issuer,
+	conn *coredata.Connector,
+) (cloud.Session, error) {
+	settings, err := coredata.ConnectorSettings[coredata.AWSConnectorSettings](conn)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read aws connector settings: %w", err)
+	}
+
+	session, err := cloudaws.NewSession(issuer, conn.OrganizationID, settings.RoleARN)
```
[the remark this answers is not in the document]
lukkor: We are OK with this. It means we changed the issuer (the application configuration) and meant to migrate these tokens anyway.

PR #1775, opened 2026-08-25
Add AWS access review

lukkor:


pkg/accessreview/drivers/aws.go
```diff
@@ -0,0 +1,157 @@
...
+	// need no account qualification and the external ID is the full user ARN.
+	AWSDriver struct {
+		account awsAccount
+	}
+
+	// awsAccount is the slice of a cloud session this driver needs. It is
+	// declared here, in the consumer, so the mapping can be exercised without
+	// an AWS endpoint: the wire protocol is covered where the calls are made,
+	// in pkg/cloud/aws.
+	awsAccount interface {
```
gearnode: k7

pkg/cloud/aws/iam.go
```diff
@@ -0,0 +1,437 @@
+// Copyright (c) 2026 Probo Inc <hello@probo.com>.
```
gearnode: should be in the driver.

PR #1784, opened 2026-08-27
Add option to skip recording GPC consent

AnotherDork:
## Problem

When a visitor's browser sends Global Privacy Control, `load()` applies the opt-out and **records it as a consent decision**: `ensureVisitorId()` writes a visitor id to `localStorage`, `setConsentCookie()` writes the consent cookie, and `recordConsent()` POSTs a consent record to the API.

That is right for a banner deployment. It is surprising in two cases:

- **Detection-only embeds.** Mounting `<probo-cookie-banner-root>` with no UI children to run the tracker detectors also makes the SDK a consent manager for GPC visitors, writing consent records the operator never asked it to write.
- **Sites with their own consent stack.** They end up with a second, narrower record of the same decision — one holding an anonymised IP and a user agent their own record may deliberately exclude.

There is currently no way to opt out of it.

## Change

- `CookieBannerClientOptions.gpcRecord?: boolean` (default `true` — unchanged behaviour), exposed as `gpc-record="false"` on `<probo-cookie-banner>` / `<probo-cookie-banner-root>` and `data-gpc-record="false"` on the IIFE script tag.
- **GPC is still honoured when the flag is off.** The opt-out is applied in the page exactly as before — `activate(gpcData)` denies non-necessary categories, blocked resources stay blocked, integrations are updated — so the flag cannot be used to ignore a GPC signal. Only the persistence and the transmission are skipped: no visitor id, no cookie, no consent record. That separation is deliberate; a flag that let operators drop the signal entirely would be the wrong thing to ship in a compliance SDK.
- The `"true"`/`"false"` attribute parsing added with `gcm-enabled` moves to a shared `resolveBooleanAttribute()` — both flags need the same warn-and-default-to-enabled behaviour, and `resolveGcmEnabled()` now delegates to it. Behaviour there is unchanged; its tests still pass untouched.
- README gains a Global Privacy Control section.

## On the shape

I have put this client-side to match `gcm-enabled`, but unlike that flag it is **not forced** to be: GPC auto-consent runs *after* the config fetch, so a per-banner `capabilities` field in the manner of `resourceReporting` would also work. I did not go that way because it is a ~25-file change across Go, GraphQL, MCP, CLI, n8n and the console locales for one boolean, and because the operator setting `gpc-record` is the same person setting `gcm-enabled` on the same embed. Happy to rework it as a server capability if you would rather have it there.

## Testing

- `npm --workspace @probo/cookie-banner run check` / `test` / `build` all pass (21 tests; new cases cover the shared resolver including the warning path).
- `gpc-record` wiring verified present in the built `dist/cookie-banner.iife.js`.
- `lint-js` steps (`@probo/skills` validate, `make relay`, root eslint + n8n lint) run locally: 0 errors.

---
## Summary by cubic
Adds an option to skip persisting and transmitting GPC consent decisions. When a visitor sends Global Privacy Control, `gpc-record="false"` (or `gpcRecord: false`) still applies the opt-out in the page — categories denied, resources blocked, integrations updated — but no longer writes a visitor id, consent cookie, or API consent record. The banner also stays hidden for those visitors, since the opt-out is in effect without a stored consent. Default behavior is unchanged.

### Tests **`+51`** **`-0`**

- Covers the new shared boolean attribute resolver: absent, `"true"`/`"false"`, case/whitespace normalization, and the warn-and-default-to-enabled path.

### Package: cookie-banner **`+87`** **`-20`**

- Adds `gpcRecord` to `CookieBannerClientOptions`, defaulting to `true` to preserve current behavior.
- When `gpcRecord` is false, `activate(gpcData)` runs directly instead of `gpc()`, so the GPC signal still takes effect without producing a record.
- Visibility now also gates on `gpcApplied`, so a GPC visitor is never shown a banner or offered categories that were already denied.
- `gpc-record` wiring is exposed on `<probo-cookie-banner>`, `<probo-cookie-banner-root>`, the themed banner, and `data-gpc-record` on the IIFE script tag.
- The `"true"`/`"false"` parsing `gcm-enabled` introduced is extracted into `resolveBooleanAttribute()`; `resolveGcmEnabled()` now delegates to it with unchanged behavior.
- README gains a Global Privacy Control section explaining the flag and that GPC remains honored.

<sup>Written for commit a1f8dcd7a1ba8bd6932325647e1dfad02787976d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1784?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


codenem (changes requested):
We will not add a parameter to deactivate the support for GPC.
Only acceptable change would be to not apply it in discovery mode, when the banner is not published

AnotherDork:
> We will not add a parameter to deactivate the support for GPC. Only acceptable change would be to not apply it in discovery mode, when the banner is not published

The opt-out suppressed only the recording; the signal was still honoured in-page (categories denied, resources blocked, integrations updated), and I fixed a banner-visibility bug so GPC visitors stayed hidden from the banner without a stored consent.
But no problem, we already managed to implement a workaround on our side that fixed the same issue. If the suppression also wont work for you, feel free to close the PR :)

PR #1792, opened 2026-08-28
Link the documented access-review connectors to their pages

aureliensibiril:
> **Merge getprobo/probo.com#94 first.** This PR adds `DocumentationURL: accessReviewDocsURL("authentik")`, and `authentik.mdx` does not exist on the docs site's `v2` branch until that PR lands. Merging this first points the console at a 404. The other four slugs already have published pages, so authentik is the only order-sensitive one.

## Why

Every documented access-review connector should carry a `DocumentationURL` so the connections list can render a link to its setup guide (`ConnectorDocumentationLink` in `AccessReviewSourceProviderListItem.tsx`, fed by `ConnectorProviderInfo.documentationUrl`). Five connectors were missing it.

## What

| Connector | Page | Was |
| --- | --- | --- |
| authentik | `/docs/product/access-review/authentik` | No page, no link |
| Brex | `/docs/product/access-review/brex` | Page existed, no link |
| Cal.com | `/docs/product/access-review/calcom` | Page existed, no link |
| Calendly | `/docs/product/access-review/calendly` | Page existed, no link |
| GitHub | `/docs/product/access-review/github` | Page existed, no link |

Nuki already had its line and needed no change. That was the other half of the problem: `nuki.go` has pointed at `/docs/product/access-review/nuki` since it shipped, but the page did not exist, so the console linked to a 404. The companion PR adds it.

After this change, every connector with a published page carries its URL, and every `DocumentationURL` in the registry resolves to a real page.

Companion docs PR: https://github.com/getprobo/probo.com/pull/94

## The test this broke

`TestAccessReviewDrivers` pinned **BREX** as its example of a provider whose `documentationUrl` is null, so documenting Brex turned that assertion into a failure:

```
FAIL: TestAccessReviewDrivers  connector_test.go:132
  Expected nil, but got: (*string)(...)
  Messages: BREX has no doc page, documentationUrl must be null
```

**AWS** replaces it. Access review for AWS is not implemented and the console still renders it as coming soon, so it cannot acquire a page by someone writing one, which is exactly how BREX stopped qualifying. The assertion now does `require.Contains` before asserting null: a bare `assert.Nil` on an absent key passes whether or not the field is really null, so the null path could have rotted unnoticed. AWS is a workload identity provider, and `base_resolvers.go` never skips those from the catalog, so the key is always present.

**Cal.com** is pinned alongside Anthropic on the positive side. Its slug is `calcom` rather than the lowercased `CAL_COM`, so it is the case that would break first if a URL were ever derived from the provider name instead of passed explicitly.

## Verification

- `go build ./pkg/connector/...` passes
- `go test ./pkg/connector/provider/` passes
- `go vet ./e2e/console/` passes, `go test -c ./e2e/console/` compiles
- `gofmt` clean
- Rebased onto `main` (`e67775aac`); CI green including `test-e2e`

## Review note

gearnode's approval was given on `a7969a20a`, which contained only the authentik line. Five of the six files here postdate it, including the whole test change this PR is about. Worth a fresh look rather than merging on that approval.

### DIFF
--- a/e2e/console/connector_test.go
+++ b/e2e/console/connector_test.go
@@ -129,7 +129,23 @@ func TestAccessReviewDrivers(t *testing.T) {
 		assert.Equal(t, "https://www.probo.com/docs/product/access-review/anthropic", *url)
 	}
 
-	assert.Nil(t, docURLByProvider["BREX"], "BREX has no doc page, documentationUrl must be null")
+	// A slug that is not the lowercased enum is the case worth pinning: Cal.com
+	// publishes at /calcom, so deriving the URL from CAL_COM would 404.
+	if url := docURLByProvider["CAL_COM"]; assert.NotNil(t, url) {
+		assert.Equal(t, "https://www.probo.com/docs/product/access-review/calcom", *url)
+	}
+
+	// AWS carries the null case because it is the one provider that cannot
+	// acquire a doc page by being written: access review for it is not
+	// implemented, and the console still renders it as coming soon. Any
+	// provider picked here purely for being undocumented today gets documented
+	// eventually and breaks this assertion, as BREX did.
+	//
+	// Contains first: a bare Nil on a missing key passes whether or not the
+	// field is really null, which would let the null path rot unnoticed. AWS is
+	// a workload identity provider, so the catalog never skips it.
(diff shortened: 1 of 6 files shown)


gearnode (approved):
LGTM

PR #1809, opened 2026-09-01
List Identity Center users in AWS access review driver

lukkor:
## Summary

- AWS access review now lists Identity Center users assigned to the connected account, including people who reach it through a group. Discovery degrades to IAM-only when the account has no instance in the session region, the role cannot read SSO Admin, or the instance lives elsewhere. There is still no Organizations walk and no cross-account assignment listing.
- Sources are named `AWS <account>` instead of the generic provider label. The worker assumes the audit role the same way the probe does, then prefers the official account name, then the sign-in alias, then the account ID. The console still shows the ID until that run finishes.

## Test plan

- [ ] Connect an account that hosts Identity Center in the session region and confirm assigned users appear (direct assignment and group membership), with unused store users omitted.
- [ ] Connect a member account, or an account whose instance is in another region, and confirm the review stays IAM-only and the fetch succeeds.
- [ ] Create two AWS sources and confirm they become `AWS <account name>` (or alias / account ID when the name is missing) and stay distinguishable.
- [ ] Confirm a source already marked name-synced keeps its current name until it is reconnected.


gearnode (approved):
LGTM

PR #1811, opened 2026-09-01
Improve identity center discovery

lukkor:
ListInstances is regional. The session stays on us-east-1 for
    STS, so an instance in another home region looked like none
    and the review dropped to IAM-only users.

Walk Identity Center regions until one returns an instance,
    then pin SSO Admin and Identity Store to that region. Empty
    or unsupported regions continue; AccessDenied still degrades
    to IAM-only.


gearnode (approved):
LGTM

PR #1828, opened 2026-09-02
Bound access source name-sync retries

aureliensibiril:
## The incident

One customer's Metabase access-review source hot-looped in production from 10:52 UTC on 2 Sep, producing **~57,000 failed attempts/hour**. At three log lines an attempt that was **81% of all prod probod log volume** (513,558 of 634,768 lines over three hours) and >99% of every WARN and ERROR the fleet emitted — every other error was buried under it.

Not a regression: `metabase.go` and `name_resolver.go` are unchanged from probod/v0.273.0 through v0.276.0. It was armed by customer data.

## Why it never stopped

- `LoadNextUnsyncedNameForUpdateSkipLocked` claims on `name_synced_at IS NULL` — a **pure read** that a failed `Process` never transitions, and whose row lock ends when the claim transaction commits.
- `kit/worker`'s inner drain loop (`worker.go:290-298`) breaks **only when `Claim` errors**. `processNext` launches `Process` in a detached goroutine and returns `nil` unconditionally (`worker.go:378`), so a `Process` failure is structurally invisible to the loop and `WithInterval(10s)` gates only the first cycle.
- Metabase answered `/api/session/properties` with **HTTP 200 and an HTML body**, so `json.Decode` failed on the first byte. `nameStatusError` classifies **status codes only** (400/401/403/404), so a decode failure was never terminal.
- Ordered by `created_at`, the poisoned row also **head-of-line blocked** name resolution for every source created after it.

This is the third recurrence of the class in this one worker — `c83770109` and `94552dbf1` ("Make Grafana/Metabase/Tailscale name errors terminal") both patched this same file for this same symptom by enumerating one more terminal case.

## How the connector got there

`doProbeRequest` rejected only 401/403 and discarded the body, so a 200 + HTML passed as a **healthy connector** at connect time. Seven providers take a customer-supplied base URL and can land on something that is not their API: Metabase, PostHog, Grafana, Authentik, Langfuse, SigNoz, Okta.

## The fix

**Retry policy moves into the claim predicate**, so it binds whatever the provider does rather than depending on a driver having named the error class. `name_sync_attempts` / `name_sync_next_attempt_at` are charged **inside the claim transaction** — the source yields its slot and `Claim` reports no task instead of spinning. Five attempts over fifteen minutes (1m/2m/4m/8m) ride out a brief outage; past that the source keeps its generic name until a reconnect clears the budget.

`ErrTerminalNameResolution` is kept, demoted to a shortcut that spares four requests on a known-permanent failure. It still earns its place — over the last 30 days it retired Notion, Tally, Brex, GitHub, Cloudflare and HubSpot failures on their first attempt — but it is no longer what guarantees termination.

Also:
- `MarkNameSynced` is a targeted update guarded on the attempt count, so a reconnect or relink landing during a resolve is no longer reverted by a full-row write from a struct read up to 10s earlier.
- The probe rejects a **2xx** whose body opens with `<`. Other statuses keep their verdict, so a provider's 5xx maintenance page stays transient rather than flipping a working connector to disconnected.
- The handler's WARN (duplicating what `kit/worker` already logs) is gone and the per-claim INFO is now Debug: 3 lines per attempt → 1.
- `worker.WithRegisterer` is wired, so `worker_tasks_total{status="failed"}` reaches a scraped registry. Its absence is why this surfaced as log volume rather than an alert.

## Behavior change

That customer's existing Metabase connector will now report **Disconnected** with `probe_failure=not_an_api_endpoint_200` instead of looking healthy — which is the point, but it is visible. Only 2xx+HTML triggers it; every probe failure in prod over the last 30 days is a genuine 401/403/OAuth rejection, so this adds no false positives to what already fails.

## Verification

- Migration + partial index applied against Postgres 17; `ADD COLUMN NOT NULL DEFAULT` is metadata-only on PG≥11 and this table holds one row per connector.
- Four `pkg/coredata` pg-harness tests pin the invariants: a charged attempt is invisible to the next claim, a backing-off source yields the queue and returns when the backoff lapses, a concurrent reset beats the worker's write, and the widened reset predicate frees a mid-backoff row. **Three of the four fail against the un-gated claim query**, so they are not vacuous.
- `go vet ./...` clean, `golangci-lint run ./pkg/...` 0 issues, all touched package tests pass.

## Operator note

This does not retroactively clear the stuck row. To stop the flood before the deploy:

```sql
UPDATE access_review_sources SET name_synced_at = now(), updated_at = now()
WHERE id = '2hR0H2lAAAEARwAAAaBhv30MjZ0pe1Ec';
```

https://claude.ai/code/session_01UYv1Yv24Zi6B7Ru6bxoHWu

### DIFF
--- a/pkg/accessreview/errors.go
+++ b/pkg/accessreview/errors.go
@@ -193,8 +193,9 @@ func (e *ProbeError) Unwrap() error {
 }
 
 // IsProviderVerdict reports whether err is the provider's answer rather than a
-// failure on Probo's side. Only a rejected credential, a transport failure that
-// reached the provider, and a refused token refresh qualify. Everything else a
+// failure on Probo's side. Only a rejected credential, a host that answered
+// with a page instead of its API, a transport failure that reached the
+// provider, and a refused token refresh qualify. Everything else a
 // probe can return (settings that will not decode, a request that could not be
 // built, a registry misconfiguration) is ours, so the default is to treat a
 // failure as Probo's and report it in full.
@@ -209,6 +210,12 @@ func IsProviderVerdict(err error) bool {
 	}
 
 	if _, ok := errors.AsType[*provider.CredentialRejectedError](err); ok {
+		return true
+	}
+
+	// The server answered, just not with its API: the URL the customer gave
+	// points somewhere else. That is their configuration, not our failure.
+	if notAPI, ok := errors.AsType[*provider.NotAnAPIEndpointError](err); ok && notAPI != nil {
 		return true
 	}
 
@@ -246,6 +253,10 @@ func ProbeFailureCode(err error) string {
 	// would cost more than the lost detail.
 	if rejected, ok := errors.AsType[*provider.CredentialRejectedError](err); ok && rejected != nil {
 		return fmt.Sprintf("credential_rejected_%d", rejected.StatusCode)
+	}
+
+	if notAPI, ok := errors.AsType[*provider.NotAnAPIEndpointError](err); ok && notAPI != nil {
+		return fmt.Sprintf("not_an_api_endpoint_%d", notAPI.StatusCode)
 	}
 
 	if _, ok := errors.AsType[*url.Error](err); ok {
--- a/pkg/accessreview/probe_error_test.go
+++ b/pkg/accessreview/probe_error_test.go
@@ -62,6 +62,15 @@ func TestProbeFailureCodeIsSafeToLog(t *testing.T) {
 	)
 	assert.Equal(t, "credential_rejected_403", accessreview.ProbeFailureCode(rejected))
 
+	// A base URL that reaches a page instead of an API is the customer's
+	// configuration, and the status is what separates it from a credential
+	// they need to reconnect. The page itself is theirs and never quoted.
+	notAPI := accessreview.NewProbeError(
+		coredata.ConnectorProviderMetabase,
+		&provider.NotAnAPIEndpointError{StatusCode: 200},
+	)
+	assert.Equal(t, "not_an_api_endpoint_200", accessreview.ProbeFailureCode(notAPI))
+
 	// A transport error embeds the customer's self-hosted host, so only the
 	// classification survives.
 	transport := accessreview.NewProbeError(
@@ -169,6 +178,7 @@ func TestIsProviderVerdict(t *testing.T) {
 
 	// The provider answered.
 	assert.True(t, accessreview.IsProviderVerdict(&provider.CredentialRejectedError{StatusCode: 401}))
+	assert.True(t, accessreview.IsProviderVerdict(&provider.NotAnAPIEndpointError{StatusCode: 200}))
 	assert.True(t, accessreview.IsProviderVerdict(&url.Error{Op: "Get", URL: "https://x.example", Err: errors.New("refused")}))
 	assert.True(t, accessreview.IsProviderVerdict(&oauth2.RetrieveError{ErrorCode: "invalid_grant"}))
 	// A workload identity connector is answered by STS through the SDK, so an
--- a/pkg/accessreview/service.go
+++ b/pkg/accessreview/service.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"go.gearno.de/kit/log"
 	"go.gearno.de/kit/pg"
 	"go.gearno.de/kit/worker"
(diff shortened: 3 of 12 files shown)


aureliensibiril:
@cubic review

aureliensibiril:
All six findings addressed in `fcd9f3b72`:

1. **`errors.go:217`** — `IsProviderVerdict`'s "Only … qualify" comment now names the non-API-endpoint branch.
2. **`access_review_source.go:725`** — right: the budget is charged at claim time, so a crash between the charge and the resolve left the row over it. `Process` now retires a row whose attempts already exceed the ceiling *before* the resolve, so a crash loop costs no further provider requests and the bound holds. Kept the ceiling in Go rather than in the claim predicate — gating there would strand an exhausted row unsynced with no terminal marker, and 4 other workers (`certificate.go:914`, `scim_bridge.go:451`, …) use the same const-plus-terminal-write shape.
3. **`probe.go:211`** — the fixed 64-byte prefix is gone; leading whitespace is now skipped across a bounded number of 512-byte reads. Covered by a `600 spaces + <!doctype` case.
4. **`probe.go:218`** — agreed, and this was the important one. A false positive retires a *working* connector, which is worse than missing a misconfiguration, so the check now matches HTML specifically (`<!doctype`/`<html`/`<head`/`<body`, case-insensitive) instead of any leading `<`. `<?xml` and a bare `<soap:Envelope` now pass; the test that enshrined the old behaviour is inverted, and `respondsWithMarkup` is renamed `respondsWithHTML`.
5. **`20260902T164733Z.sql:27`** — correct and the one with real blast radius. probod deploys blue/green (prod currently runs three versions concurrently), and the previous release's INSERT omits this column, so dropping the default would have failed those writes mid-rollout. Default retained, with a comment saying the drop belongs in a later migration. This one was self-inflicted: I added the `DROP DEFAULT` to satisfy `contrib/claude/coredata.md:467`, which doesn't account for the rollout window.
6. **`access_review_source_name_sync_test.go:143`** — the deadline is now read back against `SELECT NOW()` from the same database, so host/DB clock skew can't flake it.

Verified: full suite green with `-race` against Postgres 17, `go vet ./...` clean, `golangci-lint` 0 issues. Rebased on `40bb81e0e`.

aureliensibiril:
@cubic review

aureliensibiril:
All four addressed in `1315bb457`.

**P1 — `source_name_worker.go:300` / the reconnect race.** Correct, and it was a real hole in the guard rather than a theoretical one. Reproduced it in a test first: worker A claims (attempts 0→1) and starts resolving; a reconnect resets the budget; worker B claims and charges back to 1; A finishes and its `MarkNameSynced` matches again, landing the stale name and retiring B's claim.

Fixed by making the guard a claim generation instead of a counter — `MarkNameSynced` now also matches on `name_sync_next_attempt_at`, which `RecordNameSyncAttempt` stamps fresh from the database clock on every claim and already returned. `IS NOT DISTINCT FROM` so a NULL deadline compares correctly. No new column: the deadline is already unique per claim, which is what a generation token would have provided. `TestMarkNameSyncedYieldsToAReclaimAtTheSameAttemptCount` pins it, and fails on the stale-name and stale-retire assertions with the deadline term removed.

**P2 — budget boundary untested.** Added `TestNameSyncBudgetIsSpentOverFiveAttempts`, which claims and charges through the whole 1m/2m/4m/8m/16m schedule, asserts the row stays claimable at every step below the ceiling, and that retiring at the exhausted count takes it out of the queue. The Go-side `attempts > max` branch in `Process` is still only covered indirectly — a worker-level test needs a connector, an encryption key and a provider registry, which I did not think was worth the scaffolding here.

**P2 — `errors.go:218` nil guard.** Added. Note the branch never dereferences, so unlike `ProbeFailureCode` it could not panic — only misclassify a value no caller can construct — and the sibling `CredentialRejectedError` branch is unguarded for that reason. Taking it anyway since it is one line and you have given feedback on nil-safe probe classifiers before.

**P2 — blue/green overlap.** Agreed and now stated in the commit body: previous-release pods claim without charging an attempt, so a source failing this way keeps looping on them until the deploy drains. Marking the row synced by hand is what stops it before then — which is the operator step for the live incident regardless of this PR. I have not added a database-side compatibility mechanism; the loop is bounded by the rollout and the manual mitigation already covers the urgent case.

Verified: full suite green with `-race` against Postgres 17, `go vet ./...` clean, `golangci-lint` 0 issues. Rebased on `40bb81e0e`.

aureliensibiril:
@cubic review

PR #1837, opened 2026-09-02
Add GCP project IAM access-review driver

lukkor:
The AWS path already reviews one cloud account. GCP had a session and provider, but ListAccounts was a stub, so a connected project could not produce a source. List project IAM principals and service accounts, degrade when listing SAs is denied, and name the project from display name, then ID, then number.

---
## Summary by cubic
Adds the GCP IAM access-review driver so connected GCP projects now produce a source instead of failing on the old `ListAccounts` stub. It collects project IAM principals and service accounts, fails the fetch when either listing is denied, and names the project from display name, then ID, then project number.

### Service **`+1000`** **`-12`**

- Maps owner bindings to admin, users to SSO, and service accounts with user-managed keys to API-key authentication.
- Reports service-account status and metadata; custom roles get no admin signal rather than a false one.
- Registers the driver and a project name resolver that falls back from display name to project ID to project number.
- Adds HTTP client injection for `NewSessionFromToken` and logs Google API errors with status and reason only, keeping service-account emails out of logs.

### Tests **`+760`** **`-5`**

- Covers principal parsing, identity mapping, permission-denied failures, and project-name fallbacks with hand-authored GCP cassettes.
- Keeps cassette requests synthetic and prevents tests from using live GCP credentials.

<sup>Written for commit 405f2518cf128372a3a1ba5e7edae44db6d975f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1837?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/accessreview/drivers/gcp.go
+++ b/pkg/accessreview/drivers/gcp.go
@@ -1,0 +1,147 @@
+// Copyright (c) 2026 Probo Inc <hello@probo.com>.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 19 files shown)


lukkor:
@cubic-ai-dev review again please

pkg/accessreview/drivers/gcp.go
```diff
@@ -0,0 +1,162 @@
...
+	principals, err := listProjectIAMPrincipals(ctx, d.session)
+	if err != nil {
+		return nil, fmt.Errorf("cannot list iam identities of the gcp project: %w", err)
+	}
+
+	accounts, err := listProjectServiceAccounts(ctx, d.session)
+	if err != nil {
+		if ctx.Err() != nil {
+			return nil, err
+		}
```
gearnode: ?

PR #1847, opened 2026-09-03
Include SigNoz service accounts in access reviews

AnotherDork:
## Problem

The SigNoz access-review driver enumerates users only (`GET /api/v1/user`). SigNoz API keys belong to **service accounts**, which are a separate entity — so the credentials that authenticate automation against an instance never appear in a review, even though a service account can hold `signoz-admin`.

For anyone using the connector to evidence an access review, machine access to their observability platform is currently invisible.

## Change

Service accounts are fetched from `GET /api/v1/service_accounts` and emitted alongside users. Roles are resolved per account via `GET /api/v1/service_accounts/{id}`, because the list endpoint returns `ServiceAccount` without its roles — so admin detection works the same way it does for users, reusing the existing `sigNozRoles` normalisation.

Entries carry `AccessReviewEntryAccountTypeServiceAccount` and `AccessReviewEntryAuthMethodAPIKey`; both constants already existed, so no model change was needed.

## Degradation, deliberately

Listing service accounts requires the `serviceaccount:list` permission, which a key below `signoz-admin` does not hold. A `401`, `403` or `404` on that endpoint yields **no service accounts** rather than failing the whole source.

That keeps existing connectors working after the upgrade — they were configured before this endpoint was needed and may hold a narrower key — and leaves older instances without the route unaffected. Per-account role lookups degrade the same way: the account stays reviewable with roles unknown rather than failing the review.

## Tests

- Cassette extended with a `signoz-admin` service account and a revoked one, asserting account type, auth method, admin flag, active status and external ID.
- `TestSigNozDriverServiceAccountsForbidden` covers 401/403/404 on the list endpoint, asserting users are still returned.
- Cassette emails use synthetic `example.com` addresses to satisfy `TestCassettesUseSyntheticEmails` (SigNoz generates `<name>@signozserviceaccount.com` in practice).

Verified locally against this base: `go build`, `go test ./pkg/accessreview/...`, `gofmt -l`, and `golangci-lint run` at the CI-pinned v2.13.2 — 0 issues.

## Note for maintainers

I could not exercise the new path against a live instance: the key available to me holds `signoz-editor` + `signoz-viewer` and gets `service_account/... is not authorized to perform serviceaccount:list`. The request/response shapes are taken from SigNoz v0.137.1 (`pkg/types/serviceaccounttypes`, `pkg/apiserver/signozapiserver/serviceaccount.go`) and from a live `GET /api/v1/service_accounts/me`, which returns the same envelope with `serviceAccountRoles` populated. Worth a confirmation run with a real `signoz-admin` key before merge.

The docs page for this connector states the integration pulls user accounts only; it would need updating alongside this.

---
## Summary by cubic
The SigNoz access-review driver previously returned users only; it now includes service accounts so API-key-backed automation appears in reviews. Existing connectors continue to work when their keys cannot access service-account endpoints.

### Tests **`+53`** **`-1`**

Adds cassette-backed coverage for active and revoked accounts, denied account listing, and denied role lookups.

### Service **`+390`** **`-1`**

Fetches service accounts, resolves their roles, and records API-key authentication, account status, creation time, normalized roles, and external IDs. `401`, `403`, and `404` responses on listing return no service accounts; the same responses during role lookup leave roles and admin status unknown, while cancellations and other failures still propagate.

<sup>Written for commit 72aadcbe7040899db48a6443b54b2e986c9c7cc7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1847?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/accessreview/drivers/signoz_test.go
```diff
@@ -246,3 +246,58 @@ func TestSigNozDriverServiceAccountsForbidden(t *testing.T) {
 		})
 	}
 }
+
```
gearnode: We use k7 for testing

pkg/accessreview/drivers/signoz.go
```diff
@@ -156,15 +159,21 @@ func (d *SigNozDriver) ListAccounts(ctx context.Context) ([]AccountRecord, error
...
-		return []string{}
+		return nil, nil
 	}
 
 	assigned, err := d.queryServiceAccountRoles(ctx, id)
 	if err != nil {
-		return []string{}
+		if ctx.Err() != nil {
+			return nil, fmt.Errorf("cannot fetch signoz service account roles: %w", err)
+		}
```
gearnode: i don't get this one.

gearnode:
i do a review @AnotherDork

pkg/accessreview/drivers/signoz.go
AnotherDork: reworked it so its cleaner. The intent was that a failed role lookup shouldn't fail the whole review, but a cancelled context should.

 It now checks the errSigNozUnavailable sentinel instead, which says the same thing directly: a 401/403/404 means the roles aren't visible to this key, so they stay unknown; anything else (cancellation, network error, 5xx) propagates.

That sentinel exists because a key below signoz-admin can't call serviceaccount:list at all. Rather than fail for connectors configured before this endpoint was needed, the service-account limb degrades to empty and users are still returned.

@gearnode

AnotherDork:
@gearnode

AnotherDork:
@gearnode

AnotherDork:
@gearnode @codenem bump :3

PR #1863, opened 2026-09-07
Add purge of expired deleted files

SachaProbo:
Permanently remove S3 objects and database rows for files whose deleted_at is older than the configured retention (default 30 days). Interval, retention, and max files per tick are configurable.

---
## Summary by cubic
Adds a background worker that permanently deletes soft-deleted files from S3 and the database once they're older than the configured retention (default 30 days), instead of keeping them forever. Interval, retention, and max files per tick are configurable via env or Helm.

### Tests **`+798`** **`-0`**

Covers the new coredata queries, `DeleteFile` on S3, and purge worker behavior including S3 failures, still-referenced files, and the per-tick cap.

### Coredata **`+152`** **`-0`**

Adds queries to load soft-deleted files before a cutoff with keyset pagination, lock a deleted file with `FOR UPDATE SKIP LOCKED`, and delete a file row while returning `ErrResourceInUse` when foreign keys still reference it.

### Service **`+321`** **`-0`**

Adds the purge worker and `DeleteFile` on the S3 service, wires worker startup and shutdown into the server, and plumbs purge configuration with defaults.

### Other **`+21`** **`-0`**

Adds `.env.example` entries and Helm chart values plus deployment env vars for the purge settings.

<sup>Written for commit dbc574bdfacec2819400102f00fe4f0338219889. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1863?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/coredata/file.go
+++ b/pkg/coredata/file.go
@@ -459,3 +459,155 @@ WHERE
 
 	return nil
 }
+
+func (f *Files) LoadDeletedBefore(
+	ctx context.Context,
+	conn pg.Querier,
+	scope Scoper,
+	before time.Time,
+	afterID *gid.GID,
+	limit int,
+) error {
+	q := `
+SELECT
+    id,
+    organization_id,
+    bucket_name,
+    mime_type,
+    file_name,
+    file_key,
+    file_size,
+    visibility,
+    created_at,
+    updated_at,
+    deleted_at
+FROM
+    files
+WHERE
+    %s
+    AND deleted_at IS NOT NULL
+    AND deleted_at < @before
+    AND (
+        CASE
+            WHEN @has_after_id THEN id > @after_id
+            ELSE TRUE
+        END
+    )
+ORDER BY
+    id ASC
+LIMIT @limit
+`
+
+	q = fmt.Sprintf(q, scope.SQLFragment())
+
+	args := pgx.StrictNamedArgs{
+		"before":       before,
+		"has_after_id": afterID != nil,
+		"after_id":     gid.Nil,
+		"limit":        limit,
+	}
+	if afterID != nil {
+		args["after_id"] = *afterID
+	}
+
+	maps.Copy(args, scope.SQLArguments())
+
+	rows, err := conn.Query(ctx, q, args)
+	if err != nil {
+		return fmt.Errorf("cannot query deleted files: %w", err)
+	}
+
+	files, err := pgx.CollectRows(rows, pgx.RowToAddrOfStructByName[File])
+	if err != nil {
+		return fmt.Errorf("cannot collect deleted files: %w", err)
+	}
+
+	*f = files
+
+	return nil
+}
+
+func (f *File) LoadDeletedByIDForUpdateSkipLocked(
+	ctx context.Context,
+	conn pg.Tx,
+	scope Scoper,
+	fileID gid.GID,
+	before time.Time,
+) error {
+	q := `
+SELECT
+    id,
+    organization_id,
+    bucket_name,
+    mime_type,
+    file_name,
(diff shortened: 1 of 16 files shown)


pkg/probod/probod.go
```diff
@@ -1224,6 +1229,42 @@ func (impl *Implm) Run(
...
+		context.WithoutCancel(ctx),
+	)
+
+	wg.Go(
+		func() {
+			if err := filePurgeWorker.Run(filePurgeWorkerCtx); err != nil {
+				cancel(fmt.Errorf("file purge worker crashed: %w", err))
+			}
+		},
+	)
```
lukkor, line 1266: Open question; should we add a way to deactivate the worker? Like `if interval == 0` then the worker does not run?

pkg/filemanager/file_purge_worker.go
```diff
@@ -0,0 +1,211 @@
...
+				return fmt.Errorf("cannot lock deleted file: %w", err)
+			}
+
+			if err := locked.Delete(ctx, tx, h.scope); err != nil {
+				return fmt.Errorf("cannot delete file from database: %w", err)
+			}
+
+			if err := h.svc.DeleteFile(ctx, &locked); err != nil {
+				return fmt.Errorf("cannot delete file from S3: %w", err)
+			}
```
lukkor, line 206: I agree with the cubic comment, we should either "fire and forget" this call or do it outside the transaction.

pkg/filemanager/file_purge_worker.go
```diff
@@ -0,0 +1,211 @@
...
+			ctx,
+			"run failed",
+			log.Error(err),
+		)
+	}
+
+	return w.periodic.Run(ctx)
+}
+
+func (h *deletedFilePurgeHandler) Run(ctx context.Context) error {
```
lukkor, line 115: The run strategy here is not the right one IMO.

I would recommend one transaction loading all lines for update (skip lock) and one delete statement which returned deleted IDs then streaming those IDs outside the transaction for an async S3 delete.

We don't really "care" about a proper S3 delete as the bucket is versionned, therefore a simple delete is not enough. I think we should "fire and forget" the delete to create a delete marker and then configure the bucket to purge itself properly out of band.

pkg/coredata/file.go
```diff
@@ -459,3 +459,155 @@ WHERE
 
 	return nil
 }
+
+func (f *Files) LoadDeletedBefore(
```
gearnode, line 463: we should use Cursor

pkg/filemanager/file_purge_worker.go
```diff
@@ -0,0 +1,211 @@
...
+	DefaultDeletedFileRetention     = 30 * 24 * time.Hour
+	DefaultDeletedFileCleanupPeriod = time.Hour
+	DefaultDeletedFileCleanupBatch  = 100
+	DefaultDeletedFileMaxPerTick    = 1000
+)
+
+type (
+	DeletedFilePurgeWorker struct {
+		handler  *deletedFilePurgeHandler
+		periodic *worker.PeriodicWorker
```
gearnode, line 46: i'm not sure about this one haha we should test it ^^

SachaProbo:
I am deleting this PR as this feature a bit more refinement.

PR #1879, opened 2026-09-09
Disable cobra's mousetrap on Windows

lukkor:
Cobra aborts with a "This is a command line tool" splash and exit code 1 whenever the parent process is explorer.exe. The shell is what starts both paths Windows depends on: browsers hand probo:// URLs to explorer via IShellDispatch2::ShellExecute, and explorer owns the HKLM Run key. The last step of browser enrollment therefore died before the deep link was ever parsed, and the tray helper never came back after a logon.

Blank MousetrapHelpText, which is the switch cobra documents for this. A test guards it: package-level init runs in the test binary too, so the assertion holds on every platform.

Also add a Windows QA checklist to the release guide covering the two explorer-parented paths, since neither is reachable from a local go run.

---
## Summary by cubic
Disables cobra's mousetrap on Windows so `probo-agent` no longer aborts with a "this is a command line tool" splash and exit code 1 when its parent process is `explorer.exe`, which breaks browser enrollment and tray auto-start.

### Other **`+8`** **`-0`**

Blanks `cobra.MousetrapHelpText` in a package-level `init()` — the documented switch to disable the splash — so the deep-link handler and HKLM Run entry survive their explorer-parented launches.

<sup>Written for commit 3bf4dfc16b7d40a2ed839a4d51bbb64b3fc53a40. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1879?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmd/probo-agent/main.go
+++ b/cmd/probo-agent/main.go
@@ -54,6 +54,14 @@ var version = "dev"
 // "failed" state on a normal self-update.
 const restartExitCode = 75
 
+// Cobra prints a "this is a command line tool" splash and exits 1 when the
+// parent process is explorer.exe. The shell is what launches the probo://
+// handler and the HKLM Run entry, so the splash would break both browser
+// enrollment and tray auto-start.
+func init() {
+	cobra.MousetrapHelpText = ""
+}
+
 func main() {
 	// Best-effort cleanup of a previous-version binary left aside by
 	// a Windows self-update. No-op on Unix.


gearnode (approved):
LGTM

PR #1884, opened 2026-09-10
Add ElevenLabs, New Relic, Retool and Twingate connectors

aureliensibiril:
Four API-key, single-tenant (Pattern 3) access-review connectors, plus two
fixes to shared infrastructure that the work uncovered. Three commits, each
independently buildable.

| | |
|---|---|
| `9522d68e3` | Rewind a retried request body |
| `d4f863110` | Strip every unknown request header from cassettes |
| `1189e83a0` | The four connectors |

## Verified against the live APIs, which corrected four assumptions

Every claim about ElevenLabs, New Relic and Twingate was checked against the
real API rather than the vendor's docs. That changed the design four times:

- **ElevenLabs rejects a bad key with `400`, not `401`.** `doProbeRequest`
  reads anything but 401/403 as connected, so a dead key would have shipped a
  permanently green badge. `probeElevenLabs` adds 400 to the rejected statuses.
- **New Relic's user `type` is the billing tier, not a role.** Authority comes
  from group membership, which is why the driver walks four cursor levels and
  joins group members back onto the roster by user id.
- **New Relic has a third region, `jp`.** Supporting only us/eu would have
  locked those customers out entirely.
- **Twingate has seven roles, not four** (`HELPDESK`, `ACCESS_REVIEWER` and
  `BILLING` besides the obvious ones), confirmed by introspecting a live
  network's schema.

## Retool is Enterprise-gated

`GET /api/v2/users` needs the `users:read` scope, and that scope is absent from
the token dialog on **Team** and on **Business**. A Business token created with
every available scope selected is still refused with
`403 missing one of "users:read"`. Retool's pricing page reserves "full access
to all API scopes" for Enterprise; its internal feature catalogue carries a
`BaseEnterpriseApi` flag a Business org does not hold.

No other endpoint substitutes: `/user_invites` returns invitations rather than
members, and the usage endpoints need a `usage` scope that is also not offered.
The driver is correct and the docs state the gate. Its cassette is hand-authored
from Retool's published OpenAPI, with a header saying so and why.

## Shared fixes

**Retries dropped the POST body.** `retryRoundTripper` resent the same request
after its body had been read, so a recoverable 503 became an empty POST and a
failed sync. It affected every GraphQL driver, Monday and Railway included. The
rewind goes onto a clone, since a RoundTripper must not mutate its caller's
request.

**The cassette credential hook was fail-open.** It listed headers to remove, so
the first provider to authenticate through an unlisted name wrote its key into
a cassette verbatim — which is exactly what happened when recording ElevenLabs
(`xi-api-key` canonicalizes to `Xi-Api-Key`, not the `X-Api-Key` already
listed). Inverted to an allowlist: a missing entry now breaks a test at CI
instead of publishing a key.

## Review

Each connector was reviewed by three independent reviewers (a Claude pass, a
`review-as-gearnode` pass, and an adversarial Codex pass). They found, and this
branch fixes: silent roster truncation in New Relic's targeted pagination; a
mistyped Twingate network passing the probe as "connected"; a New Relic probe
that was green while the roster query was refused; an inverted
terminal/transient default in the name resolver; and Retool base-URL shapes
(`/api/v2/.`, `/?`) that produced wrong paths. Codex found no host-injection
bypass in Twingate's network validation.

## Testing

`go build`, the connector/accessreview/bootstrap/coredata/console-v1 suites,
`golangci-lint` (0 issues), Relay + console typecheck, and the four migrations
applied and re-applied against a disposable Postgres. All five cassettes
audited clean of credentials and real identities. New coverage includes
four-level New Relic pagination with the group-member join, Twingate cursor
paging and its missing-cursor guard, the three custom probes, and the retry
body rewind.

## Ships with

Documentation for all four in getprobo/probo.com (linked below) — these carry
`DocumentationURL`, so **the docs need to merge together with this PR** or the
four links 404.

https://claude.ai/code/session_01JFFcusrDZo9AR42g5Te2wS


aureliensibiril:
All seven findings were valid and are fixed. Notes on the three where the reasoning matters:

**P1 Twingate probe / GraphQL errors** — correct, and it was inconsistent with my own `probeNewRelic`, which already decodes the errors array. `probeTwingate` now does its own request and treats a 200-with-`errors` as the operation being refused, keeping the 404 rejection for a network that does not exist.

**P1 Retool `has_more` without a token** — correct, and the same class as the Twingate `endCursor` guard already in this branch. Both now error rather than returning a short roster: a member missing from a campaign is reviewed by nobody. Pinned by `TestRetoolDriverRefusesNextPageWithoutToken`.

**P2 Twingate rate limit** — taken rather than left documented. Pages are now spaced a second apart, waited only *between* pages, so a large network stays inside the 60-reads-a-minute budget instead of failing at roughly 6,000 users, and a single-page network waits not at all.

**P2 `Range` header** — a genuine latent regression from the allowlist inversion: no cassette carries a `Range` request header today, so nothing broke, but the next Heroku re-record would have produced a cassette that could never replay. Added to the allowlist.

**P2 New Relic transient GraphQL error** — agreed. A GraphQL errors array is now retryable rather than terminal; NerdGraph reports execution failures the same way as refusals, and the two cannot be told apart without reading provider text this driver deliberately does not read. The worker attempt budget bounds a genuine refusal anyway.

**P3 stale `Content-Length`** — fixed at the source rather than in the fixture. The sanitizers now go through `replaceCassetteBody`, which restates the length everywhere the cassette records it, and the four live cassettes were re-recorded. Three pre-existing cassettes (bitbucket, clickup, netlify) have the same inconsistency; they predate this branch and cannot be re-recorded without those credentials, so they are left alone.

**P3 dead `sanitizeAWSSigningHeaders`** — correct on both counts: the allowlist drops those headers by construction, and the AWS recorder is replay-only so no save hook ever fired. Removed along with `awsSigningHeaders`.

Verified after the fixes: `gofmt`, `go fix -diff`, `go build`, the connector/accessreview/bootstrap/coredata/console-v1 suites, and `golangci-lint` (0 issues). Each of the three commits vets in isolation.

aureliensibiril:
Second round: four of the five are fixed, one I am pushing back on with evidence.

**Rejected — `probe_test.go` "invalid JSON because its raw string contains escaped quotes".** There are no backslashes in either fixture. The literal is a Go raw string, byte-exact:

```
`{"data":null,"errors":[{"message":"x"}]}`
```

`grep | cat -A` finds zero backslashes, the string parses as JSON, and `TestProbeTwingate` passes — which it could not if `probeTwingate` were returning a decode error, since the assertion requires a `CredentialRejectedError`. No change made.

**Fixed — Twingate 404 misdiagnosed as a dead token.** Right, and it is the second time this was raised, so it should have landed the first time. A 404 now returns `NotAnAPIEndpointError`, whose classification already means "the URL the customer gave points somewhere else, that is their configuration, not our failure" — exactly this case. The type gained an optional `Detail` so it can say `no twingate network answers at this host: check the network name` instead of the markup wording, with no change for existing callers. `ProbeFailureCode` now reports `not_an_api_endpoint_404` rather than `credential_rejected_404`.

**Fixed — New Relic billing tier in `Roles`.** Agreed, and New Relic settles it: the docs call the user type a billing factor and say it is "not meant to be a way to set permissions". `Roles` on a review record means permissions, so the tier is gone rather than labelled. The docs table was corrected to match.

**Fixed — `require.*` on the stub server goroutine.** Correct and a real hang risk: `FailNow` unwinds only the handler goroutine, leaving no response written. Both `twingateStub` and `newRelicStub` now use `t.Errorf` and write an error response so the client fails fast instead of blocking.

**Partly fixed — the two-minute source deadline.** The deadline is real (`sourceFetchTimeout`), and the arithmetic is worth stating: at 60 reads a minute, two minutes buys at most 120 reads, so ~12,000 users is the hard ceiling whatever the pacing. No connector-side change can beat that.

What the flat one-second sleep did get wrong is adding the wait *on top of* each round trip. Pacing is now measured from when the previous request started, so the time Twingate spends answering counts toward the interval instead of extending it — which is what actually lets a mid-sized roster finish. The remaining ceiling is documented in the code and on the docs page as roughly ten thousand users.

Giving a rate-limited driver a longer budget than the shared default is a change to the review engine, affecting all 69 connectors, so I have left it out of a connector PR and called it out here instead. Happy to do it separately if you would rather it ship together.

Verified after the fixes: `gofmt`, `go fix -diff`, `go build`, the connector/accessreview/bootstrap/coredata/console-v1 suites, and `golangci-lint` (0 issues).

aureliensibiril:
Rebased onto main (36 commits, now level). One conflict, in `connectorSettings.ts`: main dropped the `CRISP` case when Crisp moved to the app-install connect path, and my commit added it alongside the three new ones. Resolved by dropping CRISP and keeping only `NEW_RELIC`, `RETOOL` and `TWINGATE` — the Crisp API-key setting is gone from the dialog, so there is nothing left to map.

Re-verified on the new base after regenerating: `gofmt`, `go fix -diff`, `go build`, the connector/accessreview/bootstrap/coredata/console-v1 suites, `golangci-lint` (0 issues), the Relay compiler, and the console typecheck. The registry invariants that matter after a framework change all pass against the new `installSupported` provider path — `TestEveryProviderRegistered`, `TestBuiltinRegistry_ProbeCoverage`, `TestEveryProviderSettingsReachADialog` and `TestWaveRegistrationMetadata`.

getprobo/probo.com#108 is rebased onto v2 as well, so the two still merge together.

aureliensibiril:
CI is green on the rebased branch (12/12). Two cubic findings on this run:

**1. `connector_settings.go:206` — stale "two NerdGraph endpoints" comment.** Valid, fixed (amended into `58a3842`). The branch accepts `us`, `eu` and `jp`, so the comment now says three.

**2. `newrelic_test.go:373` — PII in test failure output during a re-record run.** Not taking this one. The premise is right (the sanitizer is a `BeforeSave` hook, so a recording run asserts against live NerdGraph data), but:

- The suggested fix doesn't fix it. `assert.Contains(member.Email, "@example.com")` prints the actual string in its failure message exactly like `assert.Equal` does, so the live address still lands in the output.
- It isn't specific to this PR. Pinning the anonymized values is the suite-wide convention across ~70 driver tests (`apollo_test.go:48`, `authentik_test.go:54`, `aws_test.go:113`, `brevo_test.go:51`, `calendly_test.go:50`, …). Changing New Relic alone makes it the odd one out without removing the class of issue.
- The exposure is narrow. A recording run only happens on the machine of a developer who already holds the live `NEW_RELIC_API_KEY` and can read the roster directly; the output goes to their own terminal. CI always replays the cassette, where the values are already anonymized, so nothing reaches a shared log.

The assertions are also the thing that proves the sanitizer worked — weakening them to `NotEmpty` would let an unsanitized cassette pass. Happy to revisit as a suite-wide change if the convention should move.

PR #1886, opened 2026-09-10
Read Windows password policy without ADSI

lukkor:
WinNT://$COMPUTERNAME only exposes MinPasswordLength on workgroup hosts. Domain- and Entra-joined machines throw DISP_E_UNKNOWNNAME, so the check reported UNKNOWN.

Read the effective SAM minimum from secedit and the Intune DeviceLock PIN length from the same CSP as screen lock. Keep min_password_length in evidence so the server parser does not change.

---
## Summary by cubic
Replaces the ADSI-based Windows password policy probe, which threw DISP_E_UNKNOWNNAME on domain- and Entra-joined hosts, with reads from secedit and the MDM DeviceLock CSP. Evidence keeps `min_password_length` so the server parser is unchanged, and the `backend` field notes where the value came from.

### Coredata **`+2`** **`-0`**

- Documents that Windows reports the stricter of secedit and DeviceLock as `min_password_length`; older agents sent that key with no backend.

### Service **`+77`** **`-15`**

- Replaces the ADSI probe with a PowerShell script that exports secedit to a temp INF and reads `MinimumPasswordLength`, plus `MinDevicePasswordLength` from DeviceLock.
- Picks the stricter of the two sources and reports the backend as `max`, `secedit`, or `mdm_device_lock`.
- Reports UNKNOWN when neither source yields a password length.

### Tests **`+151`** **`-0`**

- Covers secedit INF parsing, source selection, and coredata parsing for the new evidence.

<sup>Written for commit 58be9fdf6492f1be6b6be964b7d5f25f1b9b18e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1886?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/coredata/device_posture_value_checks.go
+++ b/pkg/coredata/device_posture_value_checks.go
@@ -513,6 +513,8 @@ func parsePasswordPolicyValue(ev map[string]any) DevicePostureValue {
 		return minPasswordLengthValue(minLen)
 	}
 
+	// Windows reports the stricter of secedit and MDM DeviceLock as
+	// min_password_length. Older agents sent that key with no backend.
 	if minLen, ok := numberEvidence(ev, "min_password_length"); ok {
 		return minPasswordLengthValue(minLen)
 	}
--- a/pkg/coredata/device_posture_value_test.go
+++ b/pkg/coredata/device_posture_value_test.go
@@ -586,6 +586,26 @@ func TestParseDevicePostureValue_PasswordPolicy(t *testing.T) {
 				wantNumber: new(8),
 			},
 			{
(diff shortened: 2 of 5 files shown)


gearnode (approved):
LGTM

PR #1892, opened 2026-09-10
Keep last screen lock after Windows sign-out

lukkor:
Per-user screensaver policy disappears from the registry when nobody is signed in, so a host that already passed or failed flapped to UNKNOWN. Persist the last PASS or FAIL on the agent and reuse it for that missing observation only.

---
## Summary by cubic
The Windows screen lock check no longer flaps to UNKNOWN when nobody is signed in: per-user screensaver policy disappears from the registry at that point, so the agent now persists the last PASS or FAIL and substitutes it for that missing observation only, marking the result as remembered in the evidence.

### Service **`+151`** **`-2`**

- Adds a `check-memory.json` file on the agent that stores the last PASS or FAIL per check key, written atomically.
- Screen lock UNKNOWN results caused by no interactive user hives are now rememberable and replaced with the last stored status when one exists.
- State cleanup now also removes the check memory file.

### Tests **`+312`** **`-1`**

- Covers load/save/corrupt-file behavior, the apply/recall rules, and the full collect flow keeping the last pass on a later sign-out.
- Adds coredata parsing coverage for remembered screen lock evidence.

### Other **`+3`** **`-0`**

- Updates the agent changelog with the fix.

<sup>Written for commit 86aa6ebfc47bc0067f4ad3a4352e771ae9f237b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmd/probo-agent/CHANGELOG.md
+++ b/cmd/probo-agent/CHANGELOG.md
@@ -7,6 +7,9 @@ documented in this file.
 
 ### Fixed
 
+- The Windows screen lock check keeps the last PASS or FAIL when no
+  interactive user hives are loaded, so a signed-out host no longer flaps
+  to UNKNOWN after a user was once observed
 - The Windows screen lock check treats Entra ID users (`S-1-12-1-*`) as
   interactive hives, so a signed-in cloud account is no longer reported as
   "no interactive user hives loaded"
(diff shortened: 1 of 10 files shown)


gearnode (approved):
LGTM

PR #1915, opened 2026-09-13
Allow dots in signed cookie and token values

Ayush7614:
## What

Verify in pkg/securecookie and pkg/securetoken split the signed value on every dot and required exactly 2 parts. Any value containing a dot (emails, filenames, dotted IDs) could never round-trip through Sign then Verify.

This change splits on the last dot instead, keeping the HMAC-SHA256 construction and constant-time comparison unchanged. Current callers sign GIDs (base64url, no dots) so behavior is unchanged for them; future callers with dots now work.

## Why

Generic signed-value helpers should accept arbitrary values. The fix is small and covered by new unit tests.

## Tests

- go test -count=1 -race ./pkg/securecookie/... ./pkg/securetoken/... (pass locally)
- go vet ./pkg/securecookie/... ./pkg/securetoken/... (clean)
- gofmt -l (clean)

New tests: round-trip, values with dots, tamper rejection, wrong secret, empty secret, invalid format, cookie Get/Set round-trip, bearer Get round-trip.

Contributing: commit signed off with DCO (git commit -s).

---
## Summary by cubic
Fixes signed cookie and token verification so values containing dots can now round-trip through Sign and Verify.

### Service **`+6`** **`-6`**

Replaces splitting on every dot with splitting on the last dot, so the signed value can contain dots.

### Tests **`+293`** **`-0`**

Adds round-trip, tamper rejection, wrong secret, empty secret, and invalid format tests for both packages.

<sup>Written for commit 8870392d27a972ae3a4943dc551f344556c459e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lukkor:
Thanks for the PR — nice catch on the round-trip edge case.

We're going to close this without merging. In practice, the values we sign through `securecookie`/`securetoken` are GIDs and don't contain dots in the vast majority of cases (~99%), so the failure mode this fixes isn't one we hit. For the rare case where a dotted value did get passed in, we're fine with the current behavior of the error bubbling up rather than adding the extra split-on-last-dot logic to handle it.

Appreciate you digging into it — feel free to open something else if you spot other issues.

PR #1930, opened 2026-09-16
Set federation token nbf a minute behind iat

lukkor:
Tokens used nbf equal to iat at the current second. A verifier whose clock lags the issuer can treat that as not-yet-valid and reject a just-minted assertion. AWS and GCP still accept an nbf in the recent past.

---
## Summary by cubic
Sets federation token `nbf` one minute behind `iat` so verifiers with lagging clocks don't reject just-minted assertions. Previously `nbf` equaled `iat` at the current second, which a slow verifier could treat as not-yet-valid. AWS and GCP still accept an `nbf` in the recent past.

### Tests **`+3`** **`-1`**

Updates the claims test to expect `nbf` to be 60 seconds behind `iat`.

### Service **`+7`** **`-5`**

Sets `NotBefore` to `now.Add(-time.Minute).Unix()` in the identity federation issuer.

<sup>Written for commit 38027b654119501c8ca3820a4ad4f5bb429c1a50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


codenem (approved):

gearnode (approved):
LGTM

PR #1933, opened 2026-09-17
Strip marks from non-text nodes before saving

AnotherDork:
## Problem

Saving a document from the rich editor can fail with:

```
content cannot validate document content: hardBreak cannot have marks
```

Once it happens, every subsequent edit of that document fails with the same
error — even deleting a single space — until the page is reloaded.

## Cause

`AddMarkStep` decides whether to apply a mark by asking the **parent** node
whether it allows the mark type, not the node itself. Selecting a run of text
that spans a hard break and applying bold (bubble menu or Ctrl+B) therefore
marks the `hardBreak` node too:

```
stored doc (bold around a bare break):       0 marks on the break
plain text, select across break, bold it:    1 mark on the break
```

`prosemirror-model`'s DOM parser does the same for pasted HTML (`insertNode`
checks `top.type.allowsMarkType`), so `<strong>a<br>b</strong>` produces the
same node, and `<a href="..."></a>` produces an `image` carrying a link
mark.

`validateDocument` rejects marks on every non-text node (`rejectNonTextPayload`
in `pkg/prosemirror/schema.go`), so the update mutation fails. Since the whole
document is serialized on every update, the invalid node keeps failing every
later save; reloading discards the unsaved state and recovers.

Setting `marks: ""` on the HardBreak extension does not help: `hardBreak`
already resolves to an empty mark set, and ProseMirror does not strip
disallowed marks from a node once it carries them.

## Fix

Drop marks from every non-text node when the editor state is serialized. That
is the single point where content leaves the editor, so it covers all the ways
a mark can land on a non-text node rather than patching each origin.

## Testing

- `npm --workspace @probo/skills run validate`
- `make relay`
- `npx eslint packages/ui/src/RichEditor/RichEditor.tsx`
- `npx tsc --noEmit -p tsconfig.app.json` in `packages/ui`
- Ran the walk over a document with a bold `hardBreak`, a link-marked `image`
  and bold `text`: marks dropped from the first two, kept on the text.

---
## Summary by cubic
Fixes rich editor saves failing when a mark ends up on a non-text node (like a hard break or image), which the server rejects as invalid content. Because the whole document is serialized on every update, the invalid node keeps failing every later save until the page is reloaded and the unsaved state is discarded. Marks are now stripped from non-text nodes at serialization time, the single point where content leaves the editor.

### Package: `ui` **`+10`** **`-2`**

- Adds `stripNonTextMarks`, which deletes the `marks` property on any `JSONContent` node whose `type` isn't `text`.
- The function recurses through `node.content`, so marks are removed from nested nodes at any depth.
- Text nodes keep their marks, so formatting on text is unaffected.

<sup>Written for commit acff918933e6e0c2da9afe387b540d370bf0e6aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


SachaProbo (approved):

SachaProbo:
Thanks for your PR. We will merge your work (after rebase) this week.

PR #1936, opened 2026-09-17
Isolate Go CI caches and prune unused entries

lukkor:
The default actions/setup-go cache key ignores which job ran and only changes when go.mod does. Parallel lint, test, and build jobs then race to write one blob, and later runs keep restoring that first, stale value. Saving a new blob on every run would also make GOCACHE grow until restore is slower than compile.

Give each Go job its own cache prefix, include the run ID so a successful job always saves, and trim unused GOCACHE files before upload. Matrix legs append GOOS/GOARCH so they do not collide.

---
## Summary by cubic
Isolates Go CI caches so parallel lint, test, and build jobs stop racing over one `actions/setup-go` cache blob, and prunes unused GOCACHE entries so saved blobs stay small. Each job now gets its own cache key scoped by run, and only this run's working set is uploaded.

### Other **`+277`** **`-3`**

- New `.github/actions/setup-go` composite action installs Go with `cache: false` and manages a cache key scoped by runner, workflow, job, GOOS/GOARCH, Go version, branch, go.sum hash, and run ID, with restore keys that fall back to the default branch.
- `.github/actions/setup` now delegates Go setup to that action and exposes a `go-cache-key-prefix` input; make and release workflows pass a GOOS/GOARCH fragment so matrix legs do not collide.
- New `.github/actions/trim-gocache` action writes the job start time to state up front, then its post hook deletes GOCACHE files whose mtime predates the run, before `actions/cache` uploads the trimmed blob.
- `eslint.config.mjs` lints the new `.github/actions` JavaScript as CommonJS.
- `README.md` credits CloudX for the cache strategy.

<sup>Written for commit d5d3a4e0e47214ca19bfd56b72be0054bc64d095. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gearnode (approved):
LGTM

README.md
```diff
@@ -134,6 +134,10 @@ To report a security vulnerability, email [security@probo.com](mailto:security@p
 - [LinkedIn](https://www.linkedin.com/company/getprobo)
 - [Website](https://www.probo.com)
 
+## Thanks
```
gearnode, line 137: ❤️

PR #1941, opened 2026-09-21
Fix SCIM PUT 500 when externalId already belongs to another profile

SachaProbo:
Bridge retries the same Google user every sync because replace did not clear the existing unique external_id the way create already does.

---
## Summary by cubic
Fixes SCIM PUT returning a 500 when the incoming `externalId` already belongs to another profile, so replace transfers the ID the same way create does and Bridge stops retrying the same Google user every sync.

### Tests **`+36`** **`-0`**

- Adds an e2e test that replaces a profile with another profile's external ID and asserts the old profile no longer holds it.

### Coredata **`+9`** **`-0`**

- Converts unique-constraint violations on profile identity, external ID, and user name to `ErrResourceAlreadyExists`.

### Service **`+16`** **`-0`**

- Clears a conflicting external ID from its current profile before updating, matching create behavior.
- Maps uniqueness violations on update to a SCIM uniqueness error instead of a 500.

### Other **`+6`** **`-0`**

- Adds a changelog entry documenting the fix.

<sup>Written for commit 1145f0c6fa754aa9af485f3b62ebb18834604513. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/getprobo/probo/pull/1941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


codenem (approved):

gearnode (approved):
LGTM

